PR #1, opened 2021-06-07
[EX-3937] update files

pmareke:
-     Add LICENSE file to package copying it from the root folder
-     Create or update the README file
-     Check package.json to update information (homepage, author, license, description)
-     Check package.json to update dependencies
-     Check package.json to update scripts

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,23 @@
 # X Components Archetype
 
+[![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+[![Created by](https://img.shields.io/badge/Created%20by-Empathy.co-green)](https://www.empathy.co)
+
 Welcome to the X Components Archetype
 
 ## What is this repository for?
 
-The purpose of this repository is to be the starting point to a customer setup using the X
-Components. Please take a look at the [documentation](./docs/index.md).
+The purpose of this repository is to be the starting point to a customer setup using the X Components. Please take a look at
+the [documentation](./docs/index.md).
 
 This repository uses [Standard Version](https://github.com/conventional-changelog/standard-version)
-for the versioning. Keep this in mind when you contribute to this project. Also remember to change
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,22 +1,23 @@
...
+[![Created by](https://img.shields.io/badge/Created%20by-Empathy.co-green)](https://www.empathy.co)
+
 Welcome to the X Components Archetype
 
 ## What is this repository for?
 
-The purpose of this repository is to be the starting point to a customer setup using the X
-Components. Please take a look at the [documentation](./docs/index.md).
+The purpose of this repository is to be the starting point to a customer setup using the X Components. Please take a look at
+the [documentation](./docs/index.md).
```
tiborux, line 11: should we update the documentation? It is referring bitbucket

package.json
```diff
@@ -31,7 +35,7 @@
   "devDependencies": {
     "@cypress/webpack-preprocessor": "~5.1.1",
     "@empathybroker/deep-merge": "^1.2.1",
-    "@empathy/eslint-plugin-x": "^1.2.0",
+    "@empathyco/eslint-plugin-x": "^1.2.0",
```
LuisMartinez15, line 38: Can you move this dependency below the `@empathy/x-translations` one? So it is ordered alphabetically ;)

README.md
LuisMartinez15: We definitely should discuss about changes in the documentation. How is this project going to be used? How are clients setups/projects going to be handled/managed? Are they going to be public in github? Cause I don't think so... 

So, in the end, I think we should have a team discussion about this topic.

LuisMartinez15 (changes requested):

README.md
herrardo: Yes, we should have that discussion. The client repos are going to be in Github, but private of course. We should treat this repo as the default/preferred/mostloved/startingpoint search experience to be forked/duplicated to start an easy instrumentation.

LuisMartinez15 (approved):

tiborux (approved):

PR #9, opened 2021-06-10
fix(facets): update hierarchical facet model and replace adapter

pmareke:
EX-4135

The response from the new API changes a couple of things:

- The **hierarchical category** now it's **categoryPaths**
- The children values are now under a new property called **children**

NOW
image
BEFORE
image

In order to achieve that change we need to replace on of the mappers of the facets
image

The **EmpathyHierarchicalFacetMapper** is the one to override, after creating the new mapper we need to do the replacement and the only way I found is the following
image


davidmfempathy (dismissed):

src/components/facets/facets.vue
```diff
@@ -11,6 +11,32 @@
...
+            :filter="hierarchicalFilter"
+          >
+            <label :class="cssClasses">
+              <input @change="clickFilter" :disabled="isDisabled" type="checkbox" />
+              {{ filter.label }}
+            </label>
+          </HierarchicalFilter>
+        </Filters>
+      </BaseHeaderTogglePanel>
+    </template>
```
tajespasarela, line 38: change this to use the current Facet component as in the rest of the Facets.
Also remove any custom content in the filter, this is out of the scope of this task:

```html
    <template #categoryPaths="{ facet }">
      <Facet :facet="facet">
        <AllFilter :facet="facet" />
        <Filters #default="{ filter }" :filters="facet.filters" :animation="staggeredFadeAndSlide">
          <HierarchicalFilter :filter="filter" />
        </Filters>
      </Facet>
    </template>
```

tajespasarela (changes requested):
remove the previous `q` parameter in the request: 
```
export function customRequestMapper(
  rawRequest: SearchRequest,
  request: EmpathySearchRequest
): EmpathySearchRequest {
  return Object.assign(request, {
    query: rawRequest.query,
    q: undefined
  });
}
```

src/x-components/adapter.ts
```diff
@@ -13,5 +20,13 @@ export const adapter = new EmpathyAdapterBuilder()
...
+  .setFacetConfig({ modelName: 'HierarchicalFacet' }, 'categoryPaths')
+  .configureContainer(container => {
+    container.unbind(DEPENDENCIES.ResponseMappers.facets);
+    container.bind(DEPENDENCIES.ResponseMappers.facets).to(EmpathyFacetMapper);
+    container.bind(DEPENDENCIES.ResponseMappers.facets).to(HierarchicalFacetMapper);
+    container.bind(DEPENDENCIES.ResponseMappers.facets).to(EmpathyNumberRangeFacetMapper);
+    container.bind(DEPENDENCIES.ResponseMappers.facets).to(EmpathySimpleFacetMapper);
+  })
   .setInstance('platform')
   .build();
```
tajespasarela, line 32: Move this file to `src/adapters` and rename the folder to  `src/adapter`

tajespasarela (changes requested):

pmareke:
> q: undefined

Nice, I tried to remove it but I didn't find the way. Thanks!

src/x-components/adapter.ts
```diff
@@ -1,29 +0,0 @@
...
-    responsePaths: {
-      relatedTags: 'data.relatedtags'
-    }
-  })
-  .setFeatureConfig('nextQueries', {
-    endpoint: 'https://search.internal.test.empathy.co/query/empathy/nextqueries',
-    responsePaths: {
-      nextQueries: 'data.nextQueries'
-    }
-  })
```
tajespasarela, line 27: undo the deletion of the NQ and RT configs

tajespasarela (changes requested):

src/components/facets/facets.vue
tajespasarela: where is this change?

tajespasarela (changes requested):
There is a pending requested change in `facets.vue`

src/components/facets/facets.vue
pmareke: I did it, I promise...

PR #24, opened 2021-07-05
Feature/ex 4187 the sort dropdown is overlapped 

davidmfempathy:
# Description

Two different commits:
- main.vue: removed CSS done in the past to fix something related to the staggered fade and slide animation.
- toolbar.vue: add CollapseFromTop transition.

I forgot to put the task https://searchbroker.atlassian.net/browse/EX-4187 as the body, so please add it in the PR message.

Initially, the task describes an issue that I haven't been able to reproduce. Please, check if it happens to you but it seems that it has been fixed.

Testing it (you have to animate the sort dropdown), I noticed that something was wrong because there was a delay, pause between the start and the end of the sort animation if I use CollapseHeight. It seems to be a performance issue that doesn't happen with CollapseFromTop animation which uses scale instead of modifying the height.

# Motivation

Related issue from the original described in the task that I couldn't reproduce.

# Test

Run the server and check if the sort dropdown is overlapping any result element; check also the animation.

### DIFF
--- a/src/components/main.vue
+++ b/src/components/main.vue
@@ -179,10 +179,3 @@
     public spellcheckedQuery!: string;
   }
 </script>
-
-<style>
-  /* TODO try to find a solution to position absolute issue */
-  .x-grid .x-staggered-fade-and-slide--leave-active {
-    position: unset !important;
-  }
-</style>
--- a/src/components/toolbar.vue
+++ b/src/components/toolbar.vue
@@ -9,7 +9,12 @@
       <span v-else>{{ column }}</span>
     </ColumnPicker>
 
-    <SortDropdown v-if="$x.totalResults" class="x-dropdown--l" :items="sortValues">
+    <SortDropdown
+      v-if="$x.totalResults"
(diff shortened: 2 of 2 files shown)


CachedaCodes:
Couldn't reproduce the issue described in the task but found one that might be related.
While the dropdown is open, hovering over a result under the dropdown will result in the "Add to cart" button popping over the dropdown.

image

I would say it's similar enough to the described issue that it makes sense to handle it in this PR.

- [ ] Fix "Add to Cart button popping over dropdown"

CachedaCodes (changes requested):
Couldn't reproduce the issue described in the task but found one that might be related.
While the dropdown is open, hovering over a result under the dropdown will result in the "Add to cart" button popping over the dropdown.

image

I would say it's similar enough to the described issue that it makes sense to handle it in this PR.

- [x] Fix "Add to Cart button popping over dropdown"

CachedaCodes (approved):

tiborux (approved):

PR #26, opened 2021-07-09
feature(layout): add dir rtl to main modal

davidmfempathy:
# Description

This task was originally intended to solve a problem in which PostCSS logical, a plugin that transforms logical CSS properties to physical properties and that it's run during the build, didn't create CSS fallback rules without [dir="..."] attributes.

The library doesn't support it and the creators mention it explicitly so we have to manage and add the dir attribute directly.

# Motivation

This task came from an issue that appeared solving a task from the Archetype issues.

# Test

Switch between ltr and rtl and check that it has effect.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,6 +8,7 @@
       class="x-modal"
       modalId="x-components-app"
       :animation="collapseFromTop"
+      dir="ltr"
     >
       <Main />
     </BaseIdModal>
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,7 +8,7 @@
       class="x-modal"
       modalId="x-components-app"
       :animation="collapseFromTop"
-      dir="ltr"
+      :dir="documentDirection"
     >
       <Main />
     </BaseIdModal>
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -8,6 +8,7 @@
       class="x-modal"
       modalId="x-components-app"
       :animation="collapseFromTop"
+      dir="ltr"
```
javieri-empathy, line 11: I thought you were going to use the variable coming from the `snippetConfig`, even if I still think that thid configuration should be done at the `html` or `body` level

javieri-empathy (approved):

tiborux (approved):

src/App.vue
davidmfempathy: Done.

Agree. This configuration should be done in the html / body but we can't control what the user of the archetype does. I've upload a new approach. It gives priority to what is set in the html, then the body, then the snippetConfig and 'ltr' as fallback.

But as we've talked I'd not waste more time if we're gonna focus on the RTL in the future.

javieri-empathy (approved):

PR #31, opened 2021-08-02
build: Update TypeScript to 4.3.5 and add TSLib

tajespasarela:
EX-4480

This PR is to update the TS version to match the XComponents version.

Also fixed little naming issues in local design system tokens.

To test it, run the build and execute the project to check everything is ok


build/instrumentation.build.js
```diff
@@ -101,7 +101,6 @@ export function createConfig({
       ),
       typescript(
         mergeConfig('typescript', {
-          objectHashIgnoreUnknownHack: true,
```
pmareke, line 104: Why did you remove the parameter? Just to give some context about the change.

package.json
```diff
@@ -65,14 +67,14 @@
     "rollup-plugin-scss": "~2.1.0",
     "rollup-plugin-sourcemaps": "~0.5.0",
     "rollup-plugin-terser": "~6.1.0",
-    "rollup-plugin-typescript2": "~0.25.3",
+    "rollup-plugin-typescript2": "~0.30.0",
     "rollup-plugin-visualizer": "^4.0.4",
     "rollup-plugin-vue": "~5.1.5",
     "sass-loader": "~8.0.0",
-    "typescript": "~3.9.6",
+    "source-map-loader": "~1.1.3",
```
pmareke, line 74: Why did you add this package?

package.json
```diff
@@ -16,15 +16,16 @@
...
     "json:csv": "json-csv ./src/i18n/messages",
     "csv:json": "csv-json ./output ./src/i18n/messages"
   },
   "dependencies": {
     "@empathyco/x-adapter": "^6.2.0-alpha.4",
-    "@empathyco/x-components": "^3.0.0-alpha.9",
+    "@empathyco/x-components": "^3.0.0-alpha.10",
     "@empathyco/x-types": "^9.1.0-alpha.2",
     "reflect-metadata": "^0.1.13",
+    "tslib": "~2.3.0",
```
pmareke, line 28: Is TSLib mandatory if you use TypeScript to 4.3.5? Why are you adding this dependency now?

pmareke (approved):
Everything looks good to me, the build is okay and the app works as expected.

I left you some comments regarding the dependencies.

build/instrumentation.build.js
tajespasarela: Rollup was alerting us to change this flag. If you add it back and run the build, then you will see the warning to remove that warning.

package.json
tajespasarela: To see the source maps when we debug in Browser and we serve the project locally.

package.json
tajespasarela: It is not mandatory, but if we want to use some utils of Typescript, they are included in that library.

pmareke (approved):
Perfect!

davidmfempathy (approved):

PR #32, opened 2021-08-02
build: Replace `rollup-plugin-postcss` with `rollup-plugin-styles`.

javieri-empathy:
# Description

Replaces `rollup-plugin-postcss` with `rollup-plugin-styles`. This is done to keep a similar configurations to the x-components project.

Previous versions where throwing some warnings in the console because we were trying to transpile IE unsupported build options using autoprefixer.

# Context

This was started because the archetype was throwing some warnings when using rollup to build the project. These warnings where caused by the `grid: true` option in autoprefixer that has been removed. This option was used to transpile CSS grid to IE grid. If you bump up `rollup-plugin-postcss` version you can see the warnings that I'm talking about: 
image.

# Tests

Check that the styles are generated properly. Note that extraction is not yet fully supported. If you enable it, you should see an `styles.css` file in the output folder, but this file is not imported from the `index.html`


package.json
```diff
@@ -61,8 +61,7 @@
     "rollup-plugin-copy": "~3.4.0",
     "rollup-plugin-delete": "~1.2.0",
     "rollup-plugin-generate-html-template": "~1.7.0",
-    "rollup-plugin-postcss": "~2.0.3",
-    "rollup-plugin-scss": "~2.1.0",
+    "rollup-plugin-styles": "~3.14.1",
```
javieri-empathy, line 64: This plugin says that it requires rollup 2, but as it is working too in x-components, and we should  upgrade rollup to its latest version soon I wouldn't bother downgrading it.

package.json
```diff
@@ -61,8 +61,7 @@
     "rollup-plugin-copy": "~3.4.0",
     "rollup-plugin-delete": "~1.2.0",
     "rollup-plugin-generate-html-template": "~1.7.0",
-    "rollup-plugin-postcss": "~2.0.3",
-    "rollup-plugin-scss": "~2.1.0",
```
pmareke, line 65: The PR is about `rollup-plugin-postcss` and `rollup-plugin-styles`, why did you remove the `rollup-plugin-scss` too? Just to give some context.

pmareke (approved):
Looks good to me, just some comment regarding one deleting dependency.

build/instrumentation.build.js
```diff
@@ -113,9 +111,9 @@ export function createConfig({
           include: ['**/*.js', '**/*.mjs']
         })
       ),
-      postcss(
-        mergeConfig('postcss', {
-          extract: extractCSS,
+      styles(
+        mergeConfig('styles', {
+          mode: extractCSS ? ['extract', './style.css'] : 'inject',
```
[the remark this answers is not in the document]
tajespasarela: The link to `style.css` is not added to HTML if `extractCss` is true.

Maybe we can add this plugin: https://github.com/rollup/plugins/tree/master/packages/html#readme
to do it automátically.

tajespasarela (changes requested):

build/instrumentation.build.js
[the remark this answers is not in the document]
javieri-empathy: I didn't add it  in this PR because It was not working before. In Gluon if I remember correctly, I had to create an HTML plugin because it was not handling well the code splitting build. I don't know if that is solved and we can using it now.

package.json
javieri-empathy: Because it was not used 🤣, I don't know what it was doing there

tajespasarela (approved):

tajespasarela:
Added the auto inject of the extracted CSS into the HTML template

build/instrumentation.build.js
```diff
@@ -9,33 +9,32 @@ import path from 'path';
...
-  cssnano({ preset: ['default', { mergeLonghand: false }] })
-];
+const postCSSPlugins = [autoprefixer(), cssnano({ preset: ['default', { mergeLonghand: false }] })];
 
 /**
  * Creates a rollup configuration for projects that use X-Components. This configuration can be customized with the options object.
  *
- * @param {boolean} extractCSS - If true, the build will generate a `.css` and a `.js` file.
+ * @param {boolean} extractCss - If true, the build will generate a `.css` and a `.js` file.
+ * @param {string} outputCss - If `extractCss` is true, the build will generate a `.css` in this path.
```
pmareke, line 28: * @param {string} outputCss - If `outputCss` is true, the build will generate a `.css` in this path.

pmareke (changes requested):

pmareke (approved):
Nice job!

PR #36, opened 2021-08-10
Update jenkinsfile

pmareke:
EX-4540


Jenkinsfile
```diff
@@ -1,2 +1,3 @@
 XComponentsPipeline {
+    client = 'Archetype'
```
tajespasarela, line 2: rename `client` to `instance` to make it clearer and avoid future mistakes

tajespasarela (changes requested):

tajespasarela (approved):

PR #37, opened 2021-08-11
feat(currency): add currencyFormat injection

davidmfempathy:
# Description

- Update X-Components version.
- Add implementation of the new 'currencyFormat' injection, injecting from the App the current currency (from the reactive snippetConfig) format.
- Add initial currency configuration for EUR and USD: this is supposed to be done in an intermediate project which is not created yet.
- Modify `Result` in order to allow modifying a new slot called `price` so the format can easily be overridden.

# Test

- Check that all the requirements are fulfilled and everything work as expected.
- Check that you can modify the reactive configuration from the console: `window.initX.currency = "USD"` and see how the format changes.

### DIFF
--- a/src/components/results/result.vue
+++ b/src/components/results/result.vue
@@ -24,12 +24,7 @@
         <h2 class="x-text x-text--secondary x-ellipsis">{{ result.name }}</h2>
       </div>
 
-      <!--
-        @slot Customized price slot.
-      -->
-      <slot name="price">
-        <BaseResultCurrentPrice :result="result" class="x-text x-text--accent x-text--bold" />
-      </slot>
+      <BaseResultCurrentPrice :result="result" class="x-text x-text--accent x-text--bold" />
     </BaseResultLink>
   </article>
 </template>
--- a/src/components/results/result.vue
+++ b/src/components/results/result.vue
@@ -23,11 +23,13 @@
         <h1 class="x-small x-text--bold">{{ result.season }}</h1>
         <h2 class="x-text x-text--secondary x-ellipsis">{{ result.name }}</h2>
       </div>
-      <BaseResultCurrentPrice
-        format="i,iii.dd €"
-        :result="result"
-        class="x-text x-text--accent x-text--bold"
-      />
+
+      <!--
+        @slot Customized price slot.
+      -->
+      <slot name="price">
+        <BaseResultCurrentPrice :result="result" class="x-text x-text--accent x-text--bold" />
+      </slot>
     </BaseResultLink>
   </article>
 </template>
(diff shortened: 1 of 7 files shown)


src/components/results/result.vue
```diff
@@ -23,11 +23,13 @@
         <h1 class="x-small x-text--bold">{{ result.season }}</h1>
         <h2 class="x-text x-text--secondary x-ellipsis">{{ result.name }}</h2>
       </div>
-      <BaseResultCurrentPrice
-        format="i,iii.dd €"
-        :result="result"
-        class="x-text x-text--accent x-text--bold"
-      />
+
+      <!--
```
pmareke, line 27: Why are we adding a slot here? This is the archetype and we should not add slots right? Like the name or the season in this component.

pmareke (changes requested):

src/components/results/result.vue
davidmfempathy: I added it because one of the requirements of the story is to be able to override the currency in a component prop as CDL has if I'm not wrong.

There were no possibility to do it without this "custom" `Result` slot which is used in many different places like the Recommendations, the ResultsList, etc.

Now you can pass a currency component with the `:format` to do it.

tajespasarela: It has to be possible, but there is no need to do this here. If it is necessary for some setup/customer it will be done there.

src/currency/currencies.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * TODO: This configuration should be part of an intermediate project that would be imported by
+ * the X-Archetype.
+ */
+const currencies: Record<string, string> = {
+  EUR: 'i.iii,dd €',
+  USD: '$i,iii.dd'
+};
+
+export default currencies;
```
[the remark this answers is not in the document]
pmareke:
```javascript
export const currencies: Record<string, string> = {
  EUR: 'i.iii,dd €',
  USD: '$i,iii.dd'
};
```

src/currency/currencies.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * TODO: This configuration should be part of an intermediate project that would be imported by
+ * the X-Archetype.
+ */
+const currencies: Record<string, string> = {
```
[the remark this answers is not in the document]
pmareke: And rename it to `index.ts` inside the `currencies` folder?

src/i18n/currencies.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
pmareke, line 1: I think we can remove this comment as @tajespasarela mentioned before, it makes no sense.

pmareke (changes requested):

tajespasarela (approved):

pmareke (approved):

PR #39, opened 2021-08-18
feat: add hot swap for the lang and the searchLang

tiborux:
To test it, you can open the server and open the console. There, you can type:

window.initX.lang = "es"
window.initX.searchLang = "es"

EX-4120

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -32,7 +32,7 @@ export interface Messages {
     clear: string;
     removeLabel: string;
     title: string;
-  }
+  };
 }
 ```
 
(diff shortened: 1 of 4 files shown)


docs/i18n.md
```diff
@@ -146,17 +147,25 @@ initial bundle size. It will only be downloaded when it is needed.
 
 ### Changing the locale
 
-Because changing a locale may imply downloading new messages, a `setI18nConfiguration` method has been created
-as a helper. This method is exposed after initializing the internationalization with
-`createI18NInstance` in the [src/main.ts](../src/main.ts) file.
+There are some options in the reactive config where you can set the lang and the searchLang.
```
pmareke, line 150: There are some options in the reactive config where you can set the `lang` and the `searchLang`.

src/App.vue
```diff
@@ -63,6 +64,20 @@
     public get currencyFormat(): string {
       return currencies[this.snippetConfig.currency!];
     }
+
+    @Watch('snippetConfig.lang')
+    syncLang(lang: string): void {
+      this.$setLocale(lang);
```
pmareke, line 70: I think that if you change the `lang` you should change the `searchLang` too, as you are using the lang as searchLang in the `snippet-script.js` file.

src/vue.d.ts
```diff
@@ -0,0 +1,7 @@
+import Vue from 'vue';
```
pmareke, line 1: Remove this import as it's not used.

src/x-components/plugin.options.ts
```diff
@@ -17,8 +17,7 @@ export const installXOptions: InstallXOptions = {
       messages
     });
     vue.use(i18n);
-    (window as any).setLocale = i18n.setLocale.bind(i18n); // Only for testing
```
pmareke, line 20: I'll leave the `setLocale` in the window as it could be used in some tests.

pmareke (changes requested):

docs/i18n.md
```diff
@@ -146,17 +147,25 @@ initial bundle size. It will only be downloaded when it is needed.
...
+
+- The lang is the value used for the static messages.
```
CachedaCodes, line 152: Remove both `The` and format "lang" and "searchLang" as code.

- [ ] Remove `The` and format as code

CachedaCodes (changes requested):

src/App.vue
CachedaCodes: I understand from the `snippet-script.js` that `searchLang` is dependant on `lang` when `searchLang` doesn't have a set value, so maybe it doesn't apply here. At most I would change both if there's no `searchLang` URL parameter, but still don't see it as necessary. Up to you @tiborux

src/vue.d.ts
tiborux: if you remove it you will get this error:

image

src/x-components/plugin.options.ts
tiborux: but you can change it using window.initX.lang

pmareke: If you remove this binds, you caan't right?

pmareke (approved):

CachedaCodes (approved):

PR #41, opened 2021-09-10
feat(layout): update with SingleColumnLayout and MultiColumnMaxWidthayout

tajespasarela:
EX-4714

This PR is to update the Archetype with the new Layouts for desktop and for mobile


src/components/mobile/mobile.vue
```diff
@@ -0,0 +1,143 @@
...
+    <template #header>
+      <div
+        class="
+          x-list x-list--horizontal x-list__item--expand x-list--padding-03 x-list--align-center
+        "
+      >
+        <Close class="x-button--ghost">
+          <ArrowIcon />
+        </Close>
+        <SearchBox class="x-list__item--expand" />
```
javieri-empathy, line 12: Current searchbox  font size (`14px`) is too small for apple devices, this will trigger a zoom when focused. It needs a minimum of `16px`.

src/components/empathize.vue
```diff
@@ -139,7 +155,8 @@
       TrendingTinyIcon
     }
   })
-  export default class EmpathizeComponent extends Vue {
+  export default class PredictiveLayer extends Vue {
```
javieri-empathy, line 158: Rename also the file

javieri-empathy (changes requested):

src/components/mobile/mobile.vue
tajespasarela: yes, we will include multiple buttons, inputs, etc. sizes. But for another task/PR

javieri-empathy (approved):

PR #43, opened 2021-09-15
feat(related-tags): update endpoint

tiborux:
This updates the RTs endpoint and fix the problem where the extraParams selector doesnt change the environment.

EX-4743

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -41,7 +41,7 @@ export const adapter = new EmpathyAdapterBuilder()
     }
   })
   .setFeatureConfig('relatedTags', {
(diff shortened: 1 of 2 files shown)


pmareke (changes requested):
Just to pointing out another approach, I would use the `XOn` decorator and subscribe to the 'ExtraParamsChanged'

```javascript
@XOn('ExtraParamsChanged')
setEnv({ env }: Dictionary<unknown>): void {
  if (env) {
    adapter.setConfig({
      env: (env as 'live' | 'staging' | 'test') ?? 'live'
    });
  }
}
```

We can leave the rest of the component as it's right now.

I know that this event is going to be emitted when any extra params change but the request will be relaunch only if there is env in the extra params.

@javieri-empathy @tajespasarela what do you think?

PR #51, opened 2021-11-24
feat(sku): add new commponent and styles

pmareke:
EX-4517

[Designs](https://www.sketch.com/s/ad08dbbd-2b80-4792-80bb-aff1e1de8ec2/a/zx30ZEa)

image

### DIFF
--- a/src/components/predictive-layer.vue
+++ b/src/components/predictive-layer.vue
@@ -14,8 +14,22 @@
           : 'x-row x-row--gap-06 x-row--align-start'
       "
     >
+      <IdentifierResults
+        #default="{ identifierResult }"
+        :animation="suggestionsAnimation"
+        class="x-row__item x-row__item--span-12 x-list x-list--gap-04"
+      >
+        <BaseResultLink #default="{ result }" :result="identifierResult" class="x-suggestion">
+          <BarCodeIcon />
+          <IdentifierResult class="x-text x-text--bold" :result="result" />
+          <span class="x-text x-text--bold x-ellipsis">
+            {{ result.name }}
+          </span>
+        </BaseResultLink>
+      </IdentifierResults>
+
       <QuerySuggestions
-        v-if="$x.query.searchBox"
+        v-if="$x.query.searchBox && !$x.identifierResults"
         #suggestion-content="{ queryHTML }"
         class="x-row__item x-row__item--span-4 x-list x-list--gap-03"
         :animation="suggestionsAnimation"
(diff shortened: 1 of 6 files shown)


src/components/predictive-layer.vue
```diff
@@ -14,8 +14,22 @@
           : 'x-row x-row--gap-06 x-row--align-start'
       "
     >
+      <IdentifierResults
+        #default="{ identifierResult }"
+        :animation="suggestionsAnimation"
+        class="x-row__item x-row__item--span-12 x-list x-list--gap-04"
```
javieri-empathy, line 20: The gap should be 8px, `x-list--gap-03`, like in the `QuerySuggestions`.

image

src/components/predictive-layer.vue
```diff
@@ -14,8 +14,22 @@
...
+      >
+        <BaseResultLink #default="{ result }" :result="identifierResult" class="x-suggestion">
+          <BarCodeIcon />
+          <IdentifierResult class="x-text x-text--bold" :result="result" />
```
javieri-empathy, line 24: I would limit the results to a maximum of 5, like in the query suggestions.

src/components/predictive-layer.vue
```diff
@@ -14,8 +14,22 @@
           : 'x-row x-row--gap-06 x-row--align-start'
```
javieri-empathy, line 14: Not for this PR, but I would try to improve the readability of this template using grouped ifs, and extracting some components. It is starting to be a mess.

javieri-empathy (changes requested):

src/components/predictive-layer.vue
pmareke: image

I see 16px

src/components/predictive-layer.vue
pmareke: Me too but @tajespasarela prefers to keep like it is right now.

src/components/predictive-layer.vue
pmareke: And 12px 😆 

image

tajespasarela: javi is Right. As I always say, in these cases the designs are illustrative, but we have to keep the consistency in the setup.

so change to `gap--03`

src/components/predictive-layer.vue
tajespasarela: let's leave this way at the moment. The SKU search is a special case, to find a result or a set of them. If we "cut" that list, maybe the searched result is out of that.

We can tweak it in each customer setup, but I would leave as it is by now.

src/components/predictive-layer.vue
tajespasarela: Right, not in this PR

javieri-empathy (approved):

tajespasarela (approved):

PR #60, opened 2021-12-27
feat(next-queries): update endpoint

pmareke:
In order to test this PR you need to made a build in the mono repo, add the following code in the package.json, and install dependencies:

```json
"install:local": "npm install file:../../x/packages/x-components/empathyco-x-components-3.0.0-alpha.39.tgz file:../../x/packages/search-adapter/empathyco-x-adapter-7.0.0-alpha.8.tgz file:../../x/packages/search-types/empathyco-x-types-10.0.0-alpha.13.tgz --no-save",
```

EX-5198

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,9 +9,9 @@
       "version": "0.4.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@empathyco/x-adapter": "^7.0.0-alpha.8",
-        "@empathyco/x-components": "^3.0.0-alpha.37",
-        "@empathyco/x-types": "^10.0.0-alpha.13",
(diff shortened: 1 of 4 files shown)


diegopf (commented):
On test everything seems to be fine. However, looks like the new endpoint is not available on staging. I don't know wether this is expected or not.

pmareke:
> On test everything seems to be fine. However, looks like the new endpoint is not available on staging. I don't know wether this is expected or not.

Yeah, the same history again and again, we can't do nothing about it.

javieri-empathy (approved):

tajespasarela (approved):

PR #66, opened 2022-01-25
feat(EX-5278): build archetype in chunks

diegopf:
This PR enhances the build to produce multiple chunks instead of having one big bundle app.js with all the source code. This allows us to lazy load some of the chunks and reduces the javascript needed for the initial loading. 

Currently the build output looks like this:

Current build | Size in kB
-- | --
snippet-script.js | 1,2
app.js | 757

the build output for this PR is the following:

ES modules | Size in kB |  
-- | -- | --
snippet-script.js | 1,2 |  
main.js | 1 |  
main-[hash].js | 404 |  
search-layout-[hash].js | 0,5 | Lazy
search-layout-[hash].js | 325 | Lazy
facets-and-filters-[hash].js | 38,8 | Lazy

The main benefit is that the initial javascript requested is reduced from 757 to 404 kB. As soon as our layer is opened then the search-layout chunks will be downloaded and only when there is query from the user the facets and filter will be requested.

The current version of rollup we are using (1.x) has some problems with code splitting. For instance, it needs to create a facade chunk for the search-layout (the 0,5 kB search-layout)  even though is a dynamic chunk and not a static entry point (see [rollup issue ](https://github.com/rollup/rollup/issues/3125)). In general, we should update the rollup version in a follow up task to improve the build output.

This branch depends on a new version of x-components, the one on [this branch](https://github.com/empathyco/x/tree/feature/EX-5278-build-archetype-in-chunks)


package.json
```diff
@@ -71,7 +71,7 @@
     "rollup": "~1.31.0",
     "rollup-plugin-copy": "~3.4.0",
     "rollup-plugin-delete": "~1.2.0",
-    "rollup-plugin-generate-html-template": "~1.7.0",
+    "rollup-plugin-generate-html-template": "1.6.1",
```
diegopf, line 74: We need this version of the plugin since the latest version injects into the output index.html all the scripts tags making imposible to lazy load them.

guillei10 (approved):

guillei10 (approved):

javieri-empathy (changes requested):

src/components/desktop/desktop.vue
```diff
@@ -43,14 +43,16 @@
...
     <template #main-aside>
-      <SelectedFilters
-        v-if="$x.totalResults > 0"
-        class="x-list x-list--gap-05 x-list--padding-05 x-list--padding-bottom"
-      />
-      <Facets
-        v-if="$x.totalResults > 0"
-        class="x-list--gap-06 x-list--padding-05 x-list--padding-top"
-      />
+      <template v-if="hasQuery">
```
javieri-empathy, line 46: You can raise the `v-if` to the `<template #main-aside>` tag: `<template v-if="hasQuery" #main-aside>`

src/components/mobile/mobile.vue
```diff
@@ -48,25 +47,21 @@
...
       >
-        <BaseScroll class="x-list__item--expand">
-          <Sort class="x-list--padding-04 x-list--padding-bottom" />
-          <Facets class="x-list--gap-06 x-list--padding-04 x-list--padding-bottom" />
-        </BaseScroll>
-        <SelectedFilters
-          v-if="$x.selectedFilters.length"
-          class="x-list x-list--gap-05 x-list--padding-04 x-list--padding-bottom"
-        />
+        <template v-if="hasQuery">
```
javieri-empathy, line 55: Like with desktop, raise the `hasQuery` up to the `<template #aside>` tag

src/components/results/partial-results.vue
```diff
@@ -15,18 +15,18 @@
...
-      PartialResultsList,
-      PartialQueryButton,
-      Result: ResultComponent
+      Result: ResultComponent,
+      PartialResultsList: () => import('../search-has-query').then(m => m.PartialResultsList),
+      PartialQueryButton: () => import('../search-has-query').then(m => m.PartialQueryButton)
     }
   })
-  export default class PartialResults extends Vue {}
+  export default class PartialResults extends HasQueryMixin {}
```
javieri-empathy, line 31: This component should also be lazily loaded. Partial results is a search feature, so you shouldn't need the mixin, or lazy loading components in this file.

javieri-empathy (changes requested):

src/App.vue
```diff
@@ -5,59 +5,50 @@
     <DeviceDetector @DeviceProvided="$setLocaleDevice" :breakpoints="breakpoints" />
     <Tagging />
     <UrlHandler env="env" />
-
-    <BaseEventsModal :eventsToOpenModal="openEvents" class="x-modal" :animation="collapseFromTop">
-      <Mobile v-if="$x.device === 'mobile'" />
-      <Desktop v-else />
-    </BaseEventsModal>
+    <SearchLayout v-if="isOpen" />
```
javieri-empathy, line 8: Rename it here also to `XModal`

javieri-empathy (changes requested):
Just one last rename. 

There are some things like the scroll component, or the results-list that are in the wrong bundle, but updating Rollup seems to fix it, so we can fight it in the next PR. Well done!

src/components/main.vue
```diff
@@ -32,25 +34,23 @@
...
+  import HasQueryMixin from './has-query.mixin.vue';
 
   @Component({
     components: {
       LocationProvider,
-      NoResultsMessage,
       PartialResults,
       Recommendations,
       Redirection,
       Results,
```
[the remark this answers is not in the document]
LuisMartinez15: I agree with both `PartialResults` and `Redirection`, but I'm afraid `Results` can't be lazy loaded and should go in the main chunk, as their included in the `Top Clicked`, which appears in the initial render.

package.json
LuisMartinez15: We should take this into account when updating rollup in [EX-5336](https://searchbroker.atlassian.net/browse/EX-5336)

src/components/has-searched.mixin.vue
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
```
LuisMartinez15, line 1: You can write this `mixin` in a `.ts` file instead of a `.vue` one, as it is just a `class` and does not have neither a template nor styles, and thus remove the `<script lang="ts" />` tag. This is how we have most of the `mixins` defined in `x-components`

src/components/has-searched.mixin.vue
```diff
@@ -0,0 +1,16 @@
...
+  import { UrlParams, XOn } from '@empathyco/x-components';
+  import Vue from 'vue';
+  import Component from 'vue-class-component';
+
+  @Component
+  export default class HasSearchedMixin extends Vue {
+    protected hasSearched = false;
+    @XOn(['UserAcceptedAQuery', 'ParamsLoadedFromUrl'])
```
LuisMartinez15, line 9: Also, if you could add a blank line here, so it is easier to read 😅

src/components/search/results/partial-results.vue
```diff
@@ -15,17 +15,17 @@
 </template>
 
 <script lang="ts">
-  import { PartialResultsList, PartialQueryButton } from '@empathyco/x-components/search';
   import { BaseGrid } from '@empathyco/x-components';
   import { Component, Vue } from 'vue-property-decorator';
-  import ResultComponent from './result.vue';
+  import { PartialQueryButton, PartialResultsList } from '@empathyco/x-components/search';
+  import ResultComponent from '../../results/result.vue';
```
LuisMartinez15, line 21: I guess we can rename this to `Result` and avoid renaming it later when registering it in `line 26`? I know it is not a big deal and it was like that before, just saying 😅

src/components/x-modal.vue
```diff
@@ -0,0 +1,28 @@
...
+      Desktop
+    }
+  })
+  export default class XModal extends Vue {
+    protected openEvents = ['UserClickedOpenEventsModal', 'UserOpenXProgrammatically'];
+    protected collapseFromTop = CollapseFromTop;
+  }
+</script>
+
+<style scoped></style>
```
LuisMartinez15, line 28: Remove unused `<style />` tag.

src/components/x-modal.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <BaseEventsModal :eventsToOpenModal="openEvents" class="x-modal" :animation="collapseFromTop">
```
LuisMartinez15, line 2: You don't need to add the `class="x-modal" here, as it is already added in the `x-component`. You can see this in the resulting html:

image

LuisMartinez15 (changes requested):
Good job!

src/components/search/results/partial-results.vue
diegopf: I'm not sure about this to be honest, I still find confusing the component names on the archetype and distinguish when it is coming from x or is just a component for the archetype. Anyway, in this case I think ESLint will complain since there is already a Result on RestultComponent file.

src/components/search/facets/number-range-filters.vue
```diff
@@ -1,7 +1,7 @@
 <template>
```
javieri-empathy, line 1: Is this file even used? I can't see any usage of it 🤣.

tests/e2e/cucumber/common-steps.spec.ts
```diff
@@ -11,6 +11,7 @@ Given('an intercepted search response', () => {
 });
 
 Then('start button is clicked', () => {
+  cy.get('div.x').should('exist');
```
javieri-empathy, line 14: We try to avoid using classes in testing selectors, and we obviously completely forbid tag selectors 🤣. Add a data-test selector to the `.x` div and use it here:

```suggestion
  cy.getByDataTest('x').should('exist');
```

javieri-empathy (changes requested):

src/components/main.vue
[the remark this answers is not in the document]
javieri-empathy: I don't get you, top clicked are rendered using the `Recommendations` component, which uses the `Result` (notice it's singular). Both of `Recommendations` and `Result` components are included in the `x-modal` chunk which is fine for me... `Results` (which I would probably rename to `ResultsGrid` or with any other suffix to avoid misunderstandings) Is rendered later in the "search" chunk, just when there is a search available, uses the `Result` from the previous `x-modal` chunk.

image
image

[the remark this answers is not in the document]
LuisMartinez15: I read `Result` instead of `Results` 🤦🏿 , my bad

javieri-empathy (approved):

LuisMartinez15 (approved):

PR #70, opened 2022-02-04
build: update rollup and plugins version

tiborux:
Updates rollup and plugin versions to the lastest version. We tried to change the rollup-plugin-typescript2 for the official one but we got an error. We tried different things and read a lot of posts but the only solution that we found was to continue with the previous plugin.

EX-5336


build/instrumentation.build.js
```diff
@@ -61,8 +61,21 @@ export function createConfig({
       sourcemap: true,
       assetFileNames: '[name][extname]',
       entryFileNames: 'app.js',
```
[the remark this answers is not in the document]
javieri-empathy: Okey, this can't be done yet. The core chunk exports certain utilities, and wrapping it with the polyfills plugin in a closure will create invalid JS, as the import/export tokens must appear at the top level.
The only way I can imagine for this to work is to use the top-level await together with a dynamic import to the polyfills. This should ensure that the rest of the waits to be executed, but [this feature hasn't enough support yet](https://caniuse.com/mdn-javascript_operators_await_top_level).

So when we attack the polyfills task, we either inline them in the `core` chunk, or we have to leave another entry chunk like we do now.

javieri-empathy (changes requested):

package.json
```diff
@@ -45,10 +45,10 @@
...
     "@empathyco/x-translations": "^1.1.0-alpha.4",
     "@rollup/plugin-buble": "~0.21.3",
-    "@rollup/plugin-commonjs": "~14.0.0",
+    "@rollup/plugin-commonjs": "~21.0.1",
     "@rollup/plugin-json": "~4.1.0",
-    "@rollup/plugin-node-resolve": "~8.4.0",
-    "@rollup/plugin-replace": "~2.3.2",
+    "@rollup/plugin-node-resolve": "~13.1.3",
+    "@rollup/plugin-replace": "~3.0.1",
     "@types/autoprefixer": "~9.6.1",
```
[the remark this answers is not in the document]
tiborux: We cannot update cssnano to the latest version. If we update to this version, we need to update the vue plugin to the latest version because in the current version they are using `"postcss": "^7.0.11"` and If we update css and autoprefixer to the latest version, we need to update postcss to version 8, but vue plugin is not working with it.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2020",
```
javieri-empathy, line 3: Lets wait until X PR gets merged, but I'm afraid you are going to have to change this to es2019, and set it to es2020 in the rollup config, like @diegopf did in the other PR.

javieri-empathy: This is going to be far more complicated here, probably we will need to configure babel for webpack, as X-Components are exporting modern code

vue.config.js
```diff
@@ -18,16 +18,22 @@ module.exports = {
...
-          enforce: "pre",
-          use: ["source-map-loader"],
+          enforce: 'pre',
+          use: ['source-map-loader']
+        },
+        {
+          test: /@empathyco\/x-components/,
+          exclude: /\.css$/,
+          enforce: 'pre',
+          use: ['babel-loader']
```
javieri-empathy, line 28: This is much easier. Vue CLI already adds a babel loader for JS files. It also exposes a transpileDependencies option which you can configure with `transpileDependencies: ['@empathyco/x-components']` so it transpiles x-components.

tests/e2e/webpack.config.js
```diff
@@ -32,6 +32,12 @@ module.exports = {
             loader: 'cypress-cucumber-preprocessor/lib/featuresLoader'
           }
         ]
+      },
+      {
+        test: /@empathyco\/x-components/,
+        exclude: /\.css$/,
+        enforce: 'pre',
+        use: ['babel-loader']
       }
```
javieri-empathy, line 41: If I'm not mistaken you can simplify this config. This is only needed to transpile the `forEach` utility which is imported by the support file. There is no CSS imported, but just in case I would add the `.js` extension at the end of the regex, as it is the kind of files that babel has to transpile. Also you don't need to make this a pre-loader, a normal loader is fine.

```suggestion
      {
        test: /@empathyco\/x-components.+\.js$/,
        use: ['babel-loader']
      },```

File: babel.config.js [hunk 6]
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  presets: [['@vue/app', { useBuiltIns: 'entry' }]],
```
javieri-empathy, line 2: Isn't it enough with just transpiling the nullish coalescing operator and the optional chaining? Vue preset will try to do more transformations

```suggestion
```

javieri-empathy (changes requested):

javieri-empathy (approved):

javieri-empathy (approved):

javieri-empathy (dismissed):

javieri-empathy (approved):

LuisMartinez15 (approved):

PR #79, opened 2022-02-21
build: remove `polyfills` support

LuisMartinez15:
EX-5423

Removes unnecessary `polyfills` support.
Removes `app.js` facade chunk, which only contained an import to the `x-core` chunk, and renamed the entry file from `app.js` to `x-core.js`.

## How has this been tested?

Install `x-components` locally building it from this [pull request](https://github.com/empathyco/x/pull/341), build the `x-archetype` and check that not `polyfills` are added in the `x-core` chunk.

### DIFF
--- a/build/instrumentation.build.js
+++ b/build/instrumentation.build.js
@@ -58,7 +58,7 @@ export function createConfig({
       format: 'es',
       sourcemap: true,
       assetFileNames: '[name][extname]',
-      entryFileNames: 'x-core.js',
+      entryFileNames: 'app.js',
       chunkFileNames: chunkInfo => {
         switch (chunkInfo.name) {
           case 'x-modal':
@@ -136,5 +136,3 @@ export function createConfig({
     ]
   };
 }
-
-export default createConfig();
--- a/build/instrumentation.build.js
+++ b/build/instrumentation.build.js
@@ -14,7 +14,6 @@ import { terser } from 'rollup-plugin-terser';
 import typescript from 'rollup-plugin-typescript2';
 import visualizer from 'rollup-plugin-visualizer';
 import vue from 'rollup-plugin-vue';
-import polyFillsWrapper from '../node_modules/@empathyco/x-components/build-helpers/plugins/polyfills-wrapper.plugin';
 
 const jsOutputDirectory = path.join(process.cwd(), 'dist');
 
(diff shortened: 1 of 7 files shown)


build/instrumentation.build.js
```diff
@@ -74,7 +71,7 @@ export function createConfig({
       },
       ...output
     },
-    preserveEntrySignatures: 'strict',
+    preserveEntrySignatures: false,
```
LuisMartinez15, line 74: We've changed this property value to `false` so `rollup` does not create a `facade chunk` for our entry point, which was only doing a `dynamic import`.

For further information, you can check [rollup's docs](https://rollupjs.org/guide/en/#preserveentrysignatures).

javieri-empathy: Nice change, now I think we have a great optimised way to progressively  of loading the application

javieri-empathy (changes requested):

build/instrumentation.build.js
```diff
@@ -59,11 +58,9 @@ export function createConfig({
       format: 'es',
       sourcemap: true,
       assetFileNames: '[name][extname]',
-      entryFileNames: 'app.js',
+      entryFileNames: 'x-core.js',
```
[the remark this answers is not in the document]
tiborux: it was my fault

diegopf (changes requested):
Great job, just asking for changes until the entryFileName is app.js again😅

src/adapter/adapter.ts
```diff
@@ -16,14 +15,7 @@ import { resultMapper } from './demo-result.mapper';
...
 import { EmpathyEndpointsService } from './empathy-endpoints.service';
 
-declare global {
-  interface Window {
-    initX: SnippetConfig;
-  }
-}
-
 export const adapter = new EmpathyAdapterBuilder()
-  .setEnvironment(window.initX.env ?? 'test')
```
javieri-empathy, line 26: Great job, this was breaking one of the ways of initialising the project.  However I think this will be an issue later. Right now this keeps working because the `env-selector` component subscribes to the `ExtraParamsChanged` event, and updates the search adapter config. 
However this component won't exist in clients. So the env won't be changed. As we are already working in the adapter, I would probably just try to set this configuration in the `installExtraPlugins` callback of the `InstallXOptions`, with a big TODO reminding us to remove it when the adapter is changed to be stateless.

diegopf (approved):

PR #82, opened 2022-03-09
build: update dependencies and add dependabot config

CachedaCodes:
[EX-5441](https://searchbroker.atlassian.net/browse/EX-5441)

Updating of dependencies with vulnerabilities, also updating TypeScript and adding a `dependabot.yaml`. Vulnerabilities that this PR targets can be seen by running `npm audit` and also in the security tab > dependabot alerts of the repo.

The first approach was forcing dependency versions with an external package but in the end, we're opting for updating `vue-cli` dependencies. They were causing most of the vulnerabilities and the project is still working after the update. Most of the vulnerabilities in the dependablebot alerts should also be fixed with this approach. 

The vulnerabilities still present in the project can't be updated and were deemed not-threatening.

## Type of change
- [X] Bug fix (non-breaking change that fixes an issue)

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

Run `npm audit` to see the vulnerable dependencies before and after the changes. Present vulnerabilities should be non-threatening and the dependency causing it required by another dependency that can't be updated (ie. `rollup-plugin-vue` would require to update to `v6.0.0`, but we would require Vue 3 for it to work).


cypress.json
```diff
@@ -8,6 +8,6 @@
   "supportFile": "tests/e2e/support/index.ts",
   "screenshotOnRunFailure": false,
   "video": false,
-  "testFiles": "**/*.features",
+  "testFiles": "**/*.feature",
```
CachedaCodes, line 11: After updating to Webpack v5 via `vue-cli` dependencies using the `features` option no longer works. This makes the CI slower to up to 2 minutes. Given that this is the archetype and not the main project is somewhat assumable, but we'll have to keep an eye on it and try to find a solution in the future.

diegopf (changes requested):

tiborux (approved):

tiborux (approved):

PR #84, opened 2022-03-14
test(e2e): simplify multiple-interactions.feature tests.

javieri-empathy:
EX-5621

- Groups new related results have changed and results are different steps
- Moves the scroll down action in each group of steps after the results have changed
- Removes the Cypress assertion to check that the scroll is scrollable. If there are only a few results that do not trigger overflow this could fail.


LuisMartinez15 (approved):
LGTM. Good job!

PR #113, opened 2022-06-21
feat: Add new adapter

javieri-empathy:
EX-6362

this just adds the new adapter configured for platform.

Known issues
- Topclicked fail due to `scope` parameter. This does not happen with client endpoints
- Adapter is configured for the demo platform API. In the next PR we will add support for the API standarisation initiative.
- [There is a circular dependency in the facet](https://searchbroker.atlassian.net/browse/EX-6424).
- [Result hasDiscount value is not mapped correctly.](https://searchbroker.atlassian.net/browse/EX-6421)


tiborux (dismissed):

src/adapter/adapter.ts
```diff
@@ -1,93 +1,20 @@
...
+
+declare module '@empathyco/x-types' {
+  export interface Result {
+    season: string;
+  }
+}
+
+resultSchema.$override<EmpathyDemoPlatformResult, Partial<Result>>({
+  season: 'season'
+});
```
diegopf, line 20: What if we override the `recommendationsRequestMutableSchema` to make the topClicked work for now? Something like:

```
recommendationsRequestMutableSchema.$override<
  RecommendationsRequest,
  PlatformRecommendationsRequest
>({
  start: 'start',
  rows: 'rows',
  origin: 'origin',
  extraParams: ({ extraParams }) => {
    const extraParamsCopy = { ...extraParams };
    delete extraParamsCopy.scope;
    return extraParamsCopy;
  }
});
```

With a TODO to remove this once this issue is solved?

diegopf (changes requested):

src/adapter/adapter.ts
javieri-empathy: Good idea, didn't thought about that copy.

diegopf (approved):

tiborux:
> Launching this:
> 
> http://0.0.0.0:8080/?env=live&query=tshirt
> 
> it makes a request to:
> 
> https://api.live.empathy.co/search/v1/query/empathy/search?query=tshirt&origin=url%3Aexternal&start=0&rows=24&sort=&instance=empathy&env=live&scope=desktop&lang=en&device=mobile
> 
> which i guess is not correct as it returns a 500

because live is not a valid 'env'.  Right now what you send in the env parameter is going to be in the request and we are not checking the values, we just replace the env in the endpoints. If you wanna point to production, you have to change the snippet or remove the env parameter in the snippet

tiborux (approved):

PR #114, opened 2022-06-21
feat: update search box style and config

mnavarroespinosa:
EX-6341

### DIFF
--- a/src/components/search-box.vue
+++ b/src/components/search-box.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="x-input-group x-input-group--card">
+  <div class="x-input-group x-input-group--card x-border-color--lead">
     <SearchInput
       :placeholder="$t('searchBox.placeholder')"
       :class="$x.device === 'mobile' ? 'x-padding--left-05' : 'x-padding--left-08'"
--- a/src/components/search-box.vue
+++ b/src/components/search-box.vue
@@ -1,23 +1,36 @@
 <template>
-  <div
-    class="x-input-group"
-    :class="$x.device === 'mobile' ? 'x-input-group--card' : 'x-input-group--line'"
-  >
-    <SearchIcon />
-    <SearchInput :placeholder="$t('searchBox.placeholder')" />
-    <ClearSearchInput>{{ $t('searchBox.clear') }}</ClearSearchInput>
+  <div class="x-input-group x-input-group--card">
+    <SearchInput
+      :placeholder="$t('searchBox.placeholder')"
+      :class="$x.device === 'mobile' ? 'x-padding--left-05' : 'x-padding--left-08'"
+    />
+
+    <ClearSearchInput v-if="$x.device === 'desktop' && $x.query.searchBox" class="x-font-size--04">
+      {{ $t('searchBox.clear') }}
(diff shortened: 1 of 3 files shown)


src/components/search-box.vue
```diff
@@ -1,23 +1,36 @@
 <template>
-  <div
-    class="x-input-group"
-    :class="$x.device === 'mobile' ? 'x-input-group--card' : 'x-input-group--line'"
-  >
-    <SearchIcon />
-    <SearchInput :placeholder="$t('searchBox.placeholder')" />
-    <ClearSearchInput>{{ $t('searchBox.clear') }}</ClearSearchInput>
+  <div class="x-input-group x-input-group--card">
```
diegopf, line 2: The border color should be the lead one, I think. 
```suggestion
  <div class="x-input-group x-input-group--card x-border-color--lead">
```

src/components/search-box.vue
```diff
@@ -1,23 +1,36 @@
...
+    <SearchInput
+      :placeholder="$t('searchBox.placeholder')"
```
diegopf, line 4: I think we can remove the placeholder, at least according to figma designs

src/components/search-box.vue
```diff
@@ -1,23 +1,36 @@
...
+      :class="$x.device === 'mobile' ? 'x-padding--left-05' : 'x-padding--left-08'"
+    />
+
+    <ClearSearchInput v-if="$x.device === 'desktop' && $x.query.searchBox" class="x-font-size--04">
```
diegopf, line 8: Currently the default font weight is 300 and I think it should be 400. I think you can use the `--x-number-font-weight-input-group-default-button` token to adjust it

diegopf (changes requested):

src/components/search-box.vue
tiborux: yes, you are right

src/components/search-box.vue
```diff
@@ -1,23 +1,36 @@
...
-    class="x-input-group"
-    :class="$x.device === 'mobile' ? 'x-input-group--card' : 'x-input-group--line'"
-  >
-    <SearchIcon />
-    <SearchInput :placeholder="$t('searchBox.placeholder')" />
-    <ClearSearchInput>{{ $t('searchBox.clear') }}</ClearSearchInput>
+  <div class="x-input-group x-input-group--card">
+    <SearchInput
+      :placeholder="$t('searchBox.placeholder')"
+      :class="$x.device === 'mobile' ? 'x-padding--left-05' : 'x-padding--left-08'"
```
tiborux, line 5:
```suggestion
      :class="$x.device === 'mobile' ? 'x-padding--left-06' : 'x-padding--left-08'"
```

Should be 24px

tiborux (changes requested):

src/components/search-box.vue
tiborux: it was a mistake in the designs

src/components/search-box.vue
mnavarroespinosa: Sure, I've added it!

tiborux (approved):

diegopf (approved):

PR #116, opened 2022-06-22
feat: style and config toolbar

ajperezbau:
EX-6343


src/i18n/messages/en.messages.json
```diff
@@ -48,10 +48,10 @@
     },
     "toggleAside": {
       "hideAside": "Hide Filters",
-      "showAside": "Show Filters"
+      "showAside": "Sort & Filter"
     },
     "totalResults": {
-      "message": "{totalResults} results"
+      "message": "{totalResults} num results for <span class=\"x-text x-text--bold\">{query}</span>"
```
tiborux, line 54:
```suggestion
      "message": "{totalResults} results for <span class=\"x-text x-text--bold\">{query}</span>"
```

I think that the num word is a mistake

src/components/search/results/results.vue
```diff
@@ -3,7 +3,7 @@
     <BannersList>
       <PromotedsList>
         <NextQueriesList>
-          <component :is="grid" :columns="2" class="x-grid" :animation="staggeredFadeAndSlide">
+          <BaseVariableColumnGrid class="x-grid" :animation="staggeredFadeAndSlide">
```
tiborux, line 6: Why are you adding this change? I mean, you are configuring the toolbar

src/i18n/messages/en.messages.json
ajperezbau: Yep, it is a mistake

src/components/search/results/results.vue
ajperezbau: I added that change because now the number of columns in mobile can be changed.
If I wouldn't change this component, the ColumnPicker in mobile would not have worked.

src/components/column-picker.vue
```diff
@@ -0,0 +1,35 @@
...
+
+  @Component({
+    components: {
+      BaseColumnPickerList,
+      Grid1ColIcon,
+      Grid2ColIcon
+    }
+  })
+  export default class ColumnPicker extends Vue {
+    public get columnsValues(): number[] {
```
tiborux, line 23:
```suggestion
    public get values(): number[] {
```

src/components/column-picker.vue
```diff
@@ -0,0 +1,35 @@
...
+      return this.$x.device === 'mobile' ? [2, 1] : [4, 2];
+    }
+
+    columnIcon(column: number): unknown {
```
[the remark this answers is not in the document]
tiborux: In this case you have two ways of returning the right value:

1. return a string and you just have to wrap the name of the components as strings, like: 'Grid2ColIcon' : 'Grid1ColIcon'

2. I tried using Vue as type, but it didnt work. You can use `ComponentOptions<Vue>` and leave the code as it is.

src/components/desktop/desktop-toolbar.vue
```diff
@@ -1,44 +1,53 @@
...
-        <FiltersIcon />
-        <span v-if="isPanelOpen">{{ $t('toggleAside.hideAside') }}</span>
-        <span v-else>{{ $t('toggleAside.showAside') }}</span>
-      </template>
+      <FiltersIcon />
+      <span>{{ $t('toggleAside.showAside') }}</span>
+      <span
+        v-if="$x.selectedFilters.length"
+        class="
+          x-background--accent
```
tiborux, line 26: I think that these classes are ok for the moment, they are not perfect but we will work on the styles later. Maybe we should add what you suggested, a badge class like:

x-badge--in-line or something like that.

tiborux (changes requested):
gj

src/components/search/results/results.vue
tiborux: Ah yes, you are fucking right, sorry.

CachedaCodes (changes requested):
Is it right that the toolbar scrolls with the results?
image

From the designs it looks like it behaves like the header and stays at the top of the page
image

CachedaCodes (changes requested):

src/components/desktop/desktop-toolbar.vue
ajperezbau: Yep, with an inline variant of the badge component this would be easier and the resultant styles better.

src/components/column-picker.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <div class="x-list x-list--horizontal x-list--align-center x-list--gap-05">
+    <span class="x-uppercase">{{ $t('columnPicker.message') }}</span>
+
+    <BaseColumnPickerList v-slot="{ column }" :columns="columnsValues">
```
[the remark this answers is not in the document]
ajperezbau: Yep, you are right.
I have fixed it by setting paddings because the border of the first element is used as the separator.

tiborux (approved):

CachedaCodes (approved):

PR #119, opened 2022-06-23
feat(related-tags): add styles

tiborux:
EX-6444


diegopf (changes requested):
Please, run prettier on the tokens.scss because it will format the color codes with lowercases. Aside from that looks good!

diegopf (approved):

CachedaCodes (approved):

PR #121, opened 2022-06-23
feat: style product cards

ajperezbau:
EX-6442


diegopf (changes requested):
The results on the top clicked have the wrong styles and I think we should fix that within this task. On desktop and mobile they seem nice though!

src/design-system/tokens.scss
```diff
@@ -39,4 +44,13 @@
   --x-size-font-input-group-default: var(--x-size-font-base-m);
   --x-size-gap-input-group-default: var(--x-size-base-05);
   --x-size-height-input-group-default: var(--x-size-base-08);
+
```
tiborux, line 47: // Result
  --x-size-gap-result-default: var(--x-size-base-02);

tiborux: This is needed because the distance between the image and the title is 8px instead of 4px.

src/components/results/result.vue
```diff
@@ -11,22 +11,31 @@
...
+      class="x-result__description x-list"
+      :class="$x.device === 'mobile' ? 'x-list--gap-01' : 'x-list--gap-02'"
+      :result="result"
+    >
+      <h1 class="x-small x-uppercase">{{ result.season }}</h1>
+      <h2 class="x-text x-ellipsis" data-test="result-title">
+        {{ result.name }}
+      </h2>
+      <div class="x-list x-list--horizontal x-list--gap-04">
+        <BaseResultCurrentPrice :result="result" class="x-text x-text--bold" />
```
tiborux, line 33: should we change the format to place the € symbol next to the price?

image


format="i.iii,dd€"

src/design-system/tokens.scss
```diff
@@ -39,4 +44,13 @@
...
+  // Button
+  --x-color-background-button-tertiary: rgba(255, 255, 255, 0.85);
+  --x-color-text-button-tertiary: var(--x-color-base-lead);
+  --x-size-border-width-button-tertiary: 0;
+
+  .x-mobile {
+    --x-size-gap-grid: var(--x-size-base-03);
```
tiborux, line 54: --x-size-gap-result-default: var(--x-size-base-01);

src/components/results/result.vue
```diff
@@ -11,22 +11,31 @@
...
     </div>
 
-    <BaseResultLink class="x-result__description x-list x-list--gap-04" :result="result">
-      <div class="x-list x-list--gap-02">
-        <h1 class="x-small x-text--bold">{{ result.season }}</h1>
-        <h2 class="x-text x-text--secondary x-ellipsis" data-test="result-title">
-          {{ result.name }}
-        </h2>
+    <BaseResultLink
+      class="x-result__description x-list"
```
tiborux, line 24:
```suggestion
      class="x-result__description"
```

With this change you fix the problem with the topclicked styles.

tiborux (changes requested):
gj

tiborux:
Add this gap inside the sliding-recommendations:

<Recommendations class="x-list--gap-04">

src/components/results/result.vue
javieri-empathy: I'm not sure wether this is intentional or not, let's ask Cristina on Monday

src/design-system/tokens.scss
javieri-empathy: This also changes in mobile to 2px.

javieri-empathy (changes requested):

src/design-system/tokens.scss
ajperezbau: Yep, fixed 👌

src/components/results/result.vue
ajperezbau: But if I delete this class then the gap is not applied 🤔 
Got it working by adding the `x-list--vertical` class.

src/components/results/result.vue
```diff
@@ -11,22 +11,31 @@
...
 
-    <div class="x-result__overlay x-list x-list--horizontal">
-      <BaseAddToCart :result="result" class="x-list__item--expand">
-        <CartIcon />
+    <div v-if="$x.device === 'desktop'" class="x-result__overlay x-list x-list--horizontal">
+      <BaseAddToCart
+        :result="result"
+        class="x-list__item--expand x-button--pill x-button--tertiary x-margin--05"
+      >
         <span>{{ $t('result.addToCart') }}</span>
```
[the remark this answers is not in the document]
ajperezbau: It seems that it is not needed, removed it 👍

src/components/results/result.vue
```diff
@@ -11,22 +11,31 @@
...
-        <h1 class="x-small x-text--bold">{{ result.season }}</h1>
-        <h2 class="x-text x-text--secondary x-ellipsis" data-test="result-title">
-          {{ result.name }}
-        </h2>
+    <BaseResultLink
+      class="x-result__description x-list"
+      :class="$x.device === 'mobile' ? 'x-list--gap-01' : 'x-list--gap-02'"
+      :result="result"
+    >
+      <h1 class="x-small x-uppercase">{{ result.season }}</h1>
```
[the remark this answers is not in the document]
ajperezbau: I suppose that an `h1` was used here to give more semantics to the result information (to give more context to screen reader users, for example), as well as the `h2` in the result name.
But looking at the HTML spec, an `h1` should apply only once on the page, so ...

diegopf (dismissed):

tiborux (dismissed):

javieri-empathy (changes requested):

javieri-empathy (approved):

PR #128, opened 2022-06-24
feat: style header

ajperezbau:
EX-6448


src/App.vue
```diff
@@ -102,4 +102,10 @@
   .x-recommendations__item {
     max-width: 142px;
   }
+
+  .x-desktop {
+    .x-empathize {
+      z-index: 1;
```
tiborux, line 108: This change is needed because the related tags appear over the empathize during the animation of open and close the empathize.

src/components/desktop/desktop.vue
```diff
@@ -16,15 +25,20 @@
             </LocationProvider>
           </div>
           <LocationProvider location="predictive_layer">
-            <RelatedTags v-if="!$x.isEmpathizeOpen && $x.relatedTags.length > 0" />
+            <RelatedTags v-if="$x.relatedTags.length > 0" />
```
tiborux, line 28: The 'if' is removed because we were hidden them when we open the empathize, so there were an animation increasing and decreasing the header, then to avoid that, we leave there the related tags always

src/components/desktop/desktop.vue
```diff
@@ -1,13 +1,22 @@
...
 
         <div
-          class="x-list x-list--vertical x-list--gap-05 x-list--align-stretch x-list__item--expand"
+          class="
+            x-row__item x-row__item--span-8
+            x-list x-list--vertical x-list--gap-05 x-list--align-stretch
+          "
         >
           <div class="x-relative">
             <SearchBox />
```
tiborux, line 22: Can you change the SearchIcon inside the component to be the tiny one:

`<SearchTinyIcon class="x-icon--l" />`

src/components/desktop/desktop.vue
```diff
@@ -1,13 +1,22 @@
...
+      >
+        <div class="x-row__item x-row__item--span-2 x-list x-padding--top-04">
           <Logo />
         </div>
 
         <div
-          class="x-list x-list--vertical x-list--gap-05 x-list--align-stretch x-list__item--expand"
+          class="
+            x-row__item x-row__item--span-8
+            x-list x-list--vertical x-list--gap-05 x-list--align-stretch
```
tiborux, line 18: is it needed?
```suggestion
            x-list x-list--vertical x-list--gap-05
```

src/components/mobile/mobile.vue
```diff
@@ -3,26 +3,26 @@
     <template #header>
       <div
         class="
-          x-list x-list--horizontal x-list__item--expand x-list--padding-03 x-list--align-center
+          x-list x-list--horizontal x-list__item--expand x-list--padding-05 x-list--align-center
         "
       >
-        <Close class="x-button--ghost">
-          <ArrowLeftIcon />
+        <Close class="x-button--ghost x-padding--left-00">
```
tiborux, line 9:
```suggestion
        <Close class="x-button--ghost x-padding--00">
```

src/components/mobile/mobile.vue
```diff
@@ -3,26 +3,26 @@
     <template #header>
       <div
         class="
-          x-list x-list--horizontal x-list__item--expand x-list--padding-03 x-list--align-center
+          x-list x-list--horizontal x-list__item--expand x-list--padding-05 x-list--align-center
```
tiborux, line 6:
```suggestion
          x-list x-list--horizontal x-list__item--expand x-list--padding-05 x-list--align-center x-list--gap--03
```

tiborux (changes requested):
gj

src/App.vue
tajespasarela: move to here the `width: 100%` of line 98

src/components/desktop/desktop.vue
tajespasarela, line 28: Style the sliding pannel button: 
image

src/components/mobile/mobile.vue
tajespasarela, line 6: replace the deprecated `x-list--padding-XX` by `x-padding--XX`

src/components/mobile/mobile.vue
```diff
@@ -3,26 +3,26 @@
...
+        <Close class="x-button--ghost x-padding--left-00">
+          <ArrowLeftIcon class="x-icon--l" />
         </Close>
         <SearchBox class="x-list__item--expand" />
       </div>
     </template>
 
     <template #sub-header>
-      <div v-if="$x.relatedTags.length > 0" class="x-list x-list--padding-03">
+      <div v-if="$x.relatedTags.length > 0" class="x-list x-list--padding-05 x-padding--top-00">
```
tajespasarela, line 17: replace the deprecated x-list--padding-XX by x-padding--XX

src/components/x-modal.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <BaseEventsModal :eventsToOpenModal="openEvents" :animation="animation">
     <Mobile v-if="$x.device === 'mobile'" />
-    <Desktop v-else />
+    <Desktop v-else class="x-desktop" />
```
tajespasarela, line 4: add also the class to mobile

src/components/mobile/mobile.vue
tajespasarela: add ` x-list__item--expand` class to avoid width issues on animation when select a RT

tajespasarela (changes requested):

src/components/x-modal.vue
ajperezbau: I didn't add it because is already added in another PR and in this one is not needed.

src/components/mobile/mobile.vue
```diff
@@ -3,26 +3,26 @@
...
         <LocationProvider location="predictive_layer">
           <RelatedTags />
         </LocationProvider>
       </div>
     </template>
 
     <template #toolbar>
-      <MobileToolbar />
+      <MobileToolbar class="x-padding--05 x-padding--top-00" />
```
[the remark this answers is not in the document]
ajperezbau: Yep, that would be ideal, but I set it that way because some elements like the related tags in the grid or the partial results will occupy the whole width.

src/components/custom-sliding-panel.vue
```diff
@@ -0,0 +1,41 @@
+<template>
```
tajespasarela, line 1: This component is used only in two places. And if these two places need a different configuration in the setup, then this is useless.

I think it is better to repeat in two places the configuration than to maintain this component in future upgrades.

src/components/mobile/mobile.vue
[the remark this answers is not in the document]
tajespasarela: Don't think of this as a Setup. This is an Archetype that must be configured for any kind of setup. And in any kind of setup may be required to have different padding/spacing in different elements.

Here, in the Archetype, it is not so bad to repeat yourself and avoid abstractions. Because you don't know what will be needed in the setup tomorrow.

So it is ok as it is.

tajespasarela (changes requested):

tajespasarela (approved):

PR #142, opened 2022-07-05
fix: archetype minor tweaks

ajperezbau:
The bug of the selected filters sliding panel not working will be solved in another task because it can't be reproduced in the archetype.

EX-6554

### DIFF
--- a/src/components/predictive-layer/predictive-layer.vue
+++ b/src/components/predictive-layer/predictive-layer.vue
@@ -37,7 +37,7 @@
       >
         <div v-if="showHistoryQueries" class="x-list x-list--gap-02">
           <div v-if="!$x.query.searchBox" class="x-list x-list--horizontal x-list--align-center">
-            <h1 class="x-small x-text--bold x-text--secondary x-list__item--expand">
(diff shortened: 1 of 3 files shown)


src/components/predictive-layer/predictive-layer.vue
```diff
@@ -37,7 +37,7 @@
       >
         <div v-if="showHistoryQueries" class="x-list x-list--gap-02">
           <div v-if="!$x.query.searchBox" class="x-list x-list--horizontal x-list--align-center">
-            <h1 class="x-small x-text--bold x-text--secondary x-list__item--expand">
+            <h1 class="x-small x-text--bold x-list__item--expand">
```
mnavarroespinosa, line 40: Should we use `x-font-color--neutral-35` instead to have the same color of the design?
```suggestion
            <h1 class="x-small x-text--bold x-font-color--neutral-35 x-list__item--expand">
```

mnavarroespinosa: Oh, okay! Thanks!

mnavarroespinosa (approved):

PR #151, opened 2022-07-19
fix: Fix mobile toolbar when no env-selector, results not wrapping correctly and selected filters badge getting cut

CachedaCodes:
[EX-6625](https://searchbroker.atlassian.net/browse/EX-6625)

# Pull request template

## Motivation and context

Minor fixes for the archetype found while doing the setups for Kenay and Izas.

- [x] Open issue. If applicable, link:

## Type of change

- [x] `Main`
- [ ] Other. Specify:


src/components/search/facets/facet-selected-filters.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="selectedFilters.length" class="x-ellipsis">
+  <div v-if="selectedFilters.length">
     <span
       v-if="selectedFilters.length >= 0"
```
javieri-empathy, line 4: Do you think that we can now merge these 2 nodes into one? The `div` is adding nothing now. I can see no difference by merging them.

src/components/results/result.vue
```diff
@@ -32,7 +32,7 @@
         {{ result.name }}
       </h2>
       <span v-if="showDescription" class="x-text">{{ result.season }}</span>
-      <div class="x-list x-list--horizontal x-list--gap-04">
+      <div class="x-list x-list--horizontal x-list--wrap x-list--justify-center x-list--gap-03">
```
javieri-empathy, line 35: The gap between the result new and old price should be 12px (our 04 option). Why did you decided to change it?
image

javieri-empathy: With this change the result price is always centered, and it shouldn't
image

src/components/mobile/mobile.vue
```diff
@@ -34,7 +34,7 @@
     </template>
 
     <template #toolbar v-if="$x.query.search">
-      <MobileToolbar class="x-padding--05 x-padding--top-00" />
+      <MobileToolbar class="x-padding--left-05 x-padding--bottom-05" />
```
javieri-empathy, line 37: I'm not entirely sure that this is well done. The icon for the bigger grid cells is confusing in terms of spacing 🤣.

What I can see in figma is that the right spacing should be a little bigger than the left spacing:
image
image

And if we remove the env selector, they are now equal.  Doesn't look bad IMHO, but it does not match the designs.
image

javieri-empathy (changes requested):

LuisMartinez15 (changes requested):
The next queries list in the grid is not rendering properly (I think the css is loaded in a different order than in [the demo](x.staging.empathy.co/), so it is displayed incorrectly (see attached image).

I think we should add the utility class `x-list--horizontal` to the `next-queries-group.vue` in line 20:

```vue
      <NextQuery
        #default="{ suggestion: nextQuery }"
        :suggestion="suggestion"
        class="
          x-tag x-tag--pill
          x-list x-list--gap-03
          x-padding--bottom-03 x-padding--top-03
          x-border-color--neutral-70
          x-background--neutral-100
        "
      >
```
image

src/components/results/result.vue
CachedaCodes: > The gap between the result new and old price should be 12px (our 04 option). Why did you decided to change it? image

The change for the gap was done because when the prices wrapped the gap between them was too big. Talked again with Cristina to confirm and she gave the OK to the gap change from `-04` to `-03`.

> With this change the result price is always centered, and it shouldn't image

Mixup with Kenay, there the price is center, will fix it.

src/components/mobile/mobile.vue
CachedaCodes: Talked with Cris and we fixed it the best we could.
The thing is that the button for the column picker in the designs and in the archetype look the same but are different.

 In the designs, the column picker icon is a button with a `padding` of `8px` and a `gap` or `padding` between elements of `8px`. In the archetype, it's just a button with a `padding` of `16px`. That creates a bit of a mismatch on the right side of the last icon. Setting the `padding-right` of the toolbar to `8px` corrects the extra padding of the icon and makes it the same as the designs.

LuisMartinez15 (dismissed):
good job!

tiborux:
We are showing 0 filters selected by default

image

We should not have the badge until we select something

tiborux (approved):

LuisMartinez15 (approved):

PR #155, opened 2022-08-01
EX-6744 Change history queries confirmation dialog text

annacv:
[EX-6744](https://searchbroker.atlassian.net/browse/EX-6744)

### DIFF
--- a/src/i18n/messages/en.messages.json
+++ b/src/i18n/messages/en.messages.json
@@ -25,7 +25,7 @@
       },
       "confirmDisableModal": {
         "title":  "Disable History Queries",
-        "message":  "Disabling History Queries you also clear all the history list.",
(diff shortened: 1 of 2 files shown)


herrardo (changes requested):
I wrote the text in this way to not use "you" or "your", but then i realized that we already use that way of communicating with the customer on the My History panel, so in that case, i like better your suggestion:

https://searchbroker.atlassian.net/browse/EX-6743?focusedCommentId=156315

CachedaCodes:
I agree with @herrardo, I find `If you disable the History Queries feature, your search history will be deleted.` better.

CachedaCodes (changes requested):
I agree with @herrardo, I find `If you disable the History Queries feature, your search history will be deleted.` better.

herrardo (approved):

CachedaCodes (approved):

PR #158, opened 2022-08-08
feat: add `NextQueryPreview` carousel in results grid

LuisMartinez15:
EX-6658

Changed `components/search/results/results.vue` component used to render the results list in the grid to render a `NextQueryPreview` instead of the `NextQueriesList`.

### DIFF
--- a/src/components/desktop/desktop-next-query-preview.vue
+++ b/src/components/desktop/desktop-next-query-preview.vue
@@ -24,7 +24,7 @@
     <BaseGrid
       #default="{ item }"
       :animation="gridAnimation"
-      :columns="4"
+      :columns="maxItemsToRender"
       :items="results.slice(0, maxItemsToRender)"
       class="x-padding--00"
     >
--- a/src/components/desktop/desktop-next-query-preview.vue
+++ b/src/components/desktop/desktop-next-query-preview.vue
@@ -62,7 +62,7 @@
       Result
     }
   })
-  export default class CustomNextQueryPreview extends Vue {
+  export default class DesktopNextQueryPreview extends Vue {
     @Prop({ default: 4 })
     protected maxItemsToRender!: number;
 
--- a/src/components/desktop/desktop-next-query-preview.vue
+++ b/src/components/desktop/desktop-next-query-preview.vue
@@ -21,18 +21,17 @@
         }}
       </span>
     </div>
-    <div class="x-list x-list--horizontal x-list--justify-space-between x-list--gap-05">
-      <Result
-        v-for="result in results.slice(0, maxItemsToRender)"
-        :key="result.id"
-        :result="result"
-        :showAddToCart="false"
-        class="x-next-query-preview__result"
-      />
-    </div>
-    <BaseEventButton
-      v-if="$x.device === 'desktop'"
-      :events="clickEvents"
+    <BaseGrid
+      #default="{ item }"
+      :animation="gridAnimation"
+      :columns="4"
+      :items="results.slice(0, maxItemsToRender)"
+      class="x-padding--00"
+    >
+      <Result :result="item" />
+    </BaseGrid>
+    <NextQuery
+      :suggestion="nextQuery"
       class="
         x-tag x-tag--pill
         x-font-weight--bold
@@ -40,38 +39,37 @@
         x-padding--top-04 x-padding--bottom-04 x-padding--right-05 x-padding--left-05
         x-font-color--lead
         x-border-color--lead
-        x-background--neutral-100
       "
     >
       {{ $t('nextQueryPreview.viewResults', { totalResults }) }}
-    </BaseEventButton>
+    </NextQuery>
   </NextQueryPreview>
 </template>
 
 <script lang="ts">
   import { Component, Prop, Vue } from 'vue-property-decorator';
-  import { BaseEventButton, XEventsTypes } from '@empathyco/x-components';
-  import { NextQueryPreview } from '@empathyco/x-components/next-queries';
-  import { NextQuery } from '@empathyco/x-types';
+  import { BaseGrid, StaggeredFadeAndSlide } from '@empathyco/x-components';
+  import { NextQuery, NextQueryPreview } from '@empathyco/x-components/next-queries';
+  import { NextQuery as NextQueryModel } from '@empathyco/x-types';
   import Result from '../results/result.vue';
 
   @Component({
     components: {
-      BaseEventButton,
+      BaseGrid,
+      NextQuery,
       NextQueryPreview,
       Result
     }
   })
   export default class DesktopNextQueryPreview extends Vue {
+    // TODO: Remove this and configure NextQueryPreview after doing EX-6819.
     @Prop({ default: 4 })
     protected maxItemsToRender!: number;
 
     @Prop({ required: true })
-    protected nextQuery!: NextQuery;
+    protected nextQuery!: NextQueryModel;
 
-    protected clickEvents: Partial<XEventsTypes> = {
-      UserAcceptedAQuery: this.nextQuery.query
-    };
+    protected gridAnimation = StaggeredFadeAndSlide;
   }
 </script>
 
@@ -79,11 +77,5 @@
   .x-base-grid__next-queries-group {
     grid-column-start: 1;
     grid-column-end: -1;
-  }
-
-  .x-desktop {
-    .x-next-query-preview__result {
-      max-width: calc(25% - var(--x-size-gap-list-05));
-    }
   }
 </style>
--- a/src/components/desktop/desktop-next-query-preview.vue
+++ b/src/components/desktop/desktop-next-query-preview.vue
@@ -23,7 +23,6 @@
     </div>
     <BaseGrid
       #default="{ item }"
-      :animation="gridAnimation"
       :columns="maxItemsToRender"
       :items="results.slice(0, maxItemsToRender)"
       class="x-padding--00"
@@ -48,7 +47,7 @@
 
 <script lang="ts">
   import { Component, Prop, Vue } from 'vue-property-decorator';
-  import { BaseGrid, StaggeredFadeAndSlide } from '@empathyco/x-components';
+  import { BaseGrid } from '@empathyco/x-components';
   import { NextQuery, NextQueryPreview } from '@empathyco/x-components/next-queries';
   import { NextQuery as NextQueryModel } from '@empathyco/x-types';
   import Result from '../results/result.vue';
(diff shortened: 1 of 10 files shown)


javieri-empathy (changes requested):

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,89 @@
...
+          })
+        }}
+      </span>
+    </div>
+    <div class="x-list x-list--horizontal x-list--justify-space-between x-list--gap-05">
+      <Result
+        v-for="result in results.slice(0, maxItemsToRender)"
+        :key="result.id"
+        :result="result"
+        :showAddToCart="false"
```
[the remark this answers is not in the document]
LuisMartinez15: Sure. I'll keep it.

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,89 @@
...
+      </span>
+      <span class="x-text">
+        {{
+          $t('nextQueryPreview.totalResults', {
+            totalResults: totalResults
+          })
+        }}
+      </span>
+    </div>
+    <div class="x-list x-list--horizontal x-list--justify-space-between x-list--gap-05">
```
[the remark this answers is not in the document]
LuisMartinez15: > The description of the task said that a sliding panel should be rendered, however I can't see it. It is complex to style results width if we add it though. Can we confirm that we don't want an SlidingPanel here?

I've talked with @Alicia about the designs of this component. The `Sliding Panel` is only used in the `mobile` version, rendering an unlimited number of items. The `desktop` version, however, does not render a `Sliding Panel` and is limited to a maximum of 4 items.

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,89 @@
...
+      <Result
+        v-for="result in results.slice(0, maxItemsToRender)"
+        :key="result.id"
+        :result="result"
+        :showAddToCart="false"
+        class="x-next-query-preview__result"
+      />
+    </div>
+    <BaseEventButton
+      v-if="$x.device === 'desktop'"
```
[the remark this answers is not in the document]
LuisMartinez15: Good catch. I forgot to remove it after refactoring the initial approach.

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,89 @@
...
+  @Component({
+    components: {
+      BaseEventButton,
+      NextQueryPreview,
+      Result
+    }
+  })
+  export default class DesktopNextQueryPreview extends Vue {
+    @Prop({ default: 4 })
+    protected maxItemsToRender!: number;
```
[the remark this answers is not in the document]
LuisMartinez15: Should I create a task for it and leave a `TODO` here to remove it later? We need it now and creating a new task would already slow the process.

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,89 @@
...
+
+<style lang="scss">
+  .x-base-grid__next-queries-group {
+    grid-column-start: 1;
+    grid-column-end: -1;
+  }
+
+  .x-desktop {
+    .x-next-query-preview__result {
+      max-width: calc(25% - var(--x-size-gap-list-05));
```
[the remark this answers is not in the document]
LuisMartinez15: Sure. I'll refactor it.

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,89 @@
...
+    <div class="x-list x-list--horizontal x-list--justify-space-between x-list--gap-05">
+      <Result
+        v-for="result in results.slice(0, maxItemsToRender)"
+        :key="result.id"
+        :result="result"
+        :showAddToCart="false"
+        class="x-next-query-preview__result"
+      />
+    </div>
+    <BaseEventButton
```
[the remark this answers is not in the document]
LuisMartinez15: My bad. I'll use it.

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,95 @@
...
+      margin-left: calc(-1 * var(--x-size-padding-grid));
+
+      &__result {
+        max-width: calc(38vw - var(--x-size-gap-list-05));
+      }
+    }
+
+    .x-margin--right-negative {
+      margin-right: calc(-1 * var(--x-size-padding-grid));
+    }
```
mnavarroespinosa, line 93: As this style is not scoped, I think I would move it inside the `.x-next-query-preview` rule. And maybe I would just apply the style using the `.x-sliding-panel` selector directly. What do you think?

```
    .x-sliding-panel {
      margin-right: calc(-1 * var(--x-size-padding-grid));
    }
```

mnavarroespinosa (commented):
good job!

src/components/desktop/desktop-next-query-preview.vue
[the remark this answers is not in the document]
LuisMartinez15: > Because of the justify space between, when there are a few items it spaces them:

This issue should be resolved after the refactor using a `BaseGrid`.

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <NextQueryPreview
```
javieri-empathy, line 2: The list of NQs does not take the full width available like in the designs.

image
VS
image

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,90 @@
...
+    class="
+      x-list x-list--gap-05
+      x-background--neutral-95
+      x-padding--06 x-padding--bottom-06
+      x-margin--top-10 x-margin--bottom-05
+    "
+    :suggestion="nextQuery"
+  >
+    <span class="x-title2 x-font-weight--bold">{{ $t('nextQueryPreview.title') }}</span>
+    <div class="x-list x-list--horizontal x-list--align-center x-list--justify-space-between">
```
javieri-empathy, line 13: You don't need this `justify-space-between` utility, you can add the `margin-left-auto` one to the view all button.

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,90 @@
...
+      <div class="x-list x-list--horizontal x-list--align-center x-list--gap-02">
```
javieri-empathy, line 14: It is just 1 or 2px effect, but these items should be aligned to the baseline.

```suggestion
      <div class="x-list x-list--horizontal x-list--align-baseline x-list--gap-02">
```

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,90 @@
...
+</script>
+
+<style lang="scss">
+  .x-mobile {
+    .x-next-query-preview {
+      margin-right: calc(-1 * var(--x-size-padding-grid));
+      margin-left: calc(-1 * var(--x-size-padding-grid));
+
+      &__result {
+        max-width: calc(38vw - var(--x-size-gap-list-05));
```
javieri-empathy, line 82: I would say this is not very safe, if somehow the title/image are smaller than this size, that result would have a different width. Probably the safest way is just setting the width?

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,90 @@
...
+  export default class MobileNextQueryPreview extends Vue {
+    @Prop({ required: true })
+    protected nextQuery!: NextQueryModel;
+  }
+</script>
+
+<style lang="scss">
+  .x-mobile {
+    .x-next-query-preview {
+      margin-right: calc(-1 * var(--x-size-padding-grid));
```
javieri-empathy, line 78: Use logical properties: `margin-inline` for this case

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,81 @@
...
+
+    @Prop({ required: true })
+    protected nextQuery!: NextQueryModel;
+
+    protected gridAnimation = StaggeredFadeAndSlide;
+  }
+</script>
+
+<style lang="scss">
+  .x-base-grid__next-queries-group {
```
javieri-empathy, line 77: should these styles be in the base project too? 
In any case as we are probably going to leave them in the archetype, I don't think this is the best place, as they are affecting both desktop andmobile versions of this component

src/components/desktop/desktop-next-query-preview.vue
```diff
@@ -0,0 +1,81 @@
...
+  })
+  export default class DesktopNextQueryPreview extends Vue {
+    // TODO: Remove this and configure NextQueryPreview after doing EX-6819.
+    @Prop({ default: 4 })
+    protected maxItemsToRender!: number;
+
+    @Prop({ required: true })
+    protected nextQuery!: NextQueryModel;
+
+    protected gridAnimation = StaggeredFadeAndSlide;
```
javieri-empathy, line 72: I love animations, but is it animating properly to you? I guess this is happening at the same time than the parent grid animation, so it won't look properly. If that's the case we can remove it for now.

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -0,0 +1,90 @@
...
+          {{
+            $t('nextQueryPreview.totalResults', {
+              totalResults: totalResults
+            })
+          }}
+        </span>
+      </div>
+      <NextQuery
+        :suggestion="nextQuery"
+        class="x-button x-button--ghost x-border-width--00 x-padding--00 x-font-weight--bold"
```
javieri-empathy, line 32: Is the border-width class necessary?

src/components/mobile/mobile-next-query-preview.vue
LuisMartinez15: Sure. Done!

src/components/mobile/mobile-next-query-preview.vue
javieri-empathy: I think this is because of the wrong margin/paddings

You can set a negative margin of `24px` (The 06 value of our design system sizes) to match parent containers 24px padding, and then set a padding left of 24px too in the scroll element of the sliding panel

javieri-empathy (changes requested):

src/components/desktop/desktop-next-query-preview.vue
LuisMartinez15: I've moved this to the `results.vue` component where this slot is defined.

src/components/mobile/mobile-next-query-preview.vue
LuisMartinez15: You're right.

src/components/mobile/mobile-next-query-preview.vue
LuisMartinez15: Which browser are you using? I see it taking the full width

image

mnavarroespinosa (approved):

javieri-empathy (approved):

mnavarroespinosa (approved):

PR #170, opened 2022-09-09
fix: `API` `live` `env` should work properly now, `test` is no longer accessible

LuisMartinez15:
EX-6723


mnavarroespinosa (approved):

ajperezbau (approved):
Nice 👌

herrardo (approved):

PR #171, opened 2022-09-13
feat(queries-preview): add next query preview in pre search

ajperezbau:
EX-6926


herrardo (changes requested):
Great job!! It looks great.

There is just one part of the development missing, I think. There was info warn in the task saying:

`Currently the Archetype set focus on the Search Box right after the shopper clicks on Start, I think to show this feature, we should avoid the autofocus, but I don’t know if that should be the standard for the Archetype yet.`

So i think we should deactivate the autofocus.

Ideally, we should do autofocus when this feature is not present. But as the archetype default layout will have the feature we can just deactivate it. Then, when we do a customer setup, we should figure out a way to activate the autofocus again if the customer is not using this feature.

herrardo:
> > Great job!! It looks great.
> > There is just one part of the development missing, I think. There was info warn in the task saying:
> > `Currently the Archetype set focus on the Search Box right after the shopper clicks on Start, I think to show this feature, we should avoid the autofocus, but I don’t know if that should be the standard for the Archetype yet.`
> > So i think we should deactivate the autofocus.
> > Ideally, we should do autofocus when this feature is not present. But as the archetype default layout will have the feature we can just deactivate it. Then, when we do a customer setup, we should figure out a way to activate the autofocus again if the customer is not using this feature.
> 
> F, I missed that 🤦‍♂️ Ok! I will take a look to it right now.

no blame at all ;)

src/App.vue
```diff
@@ -59,6 +65,11 @@
       return currencies[this.snippetConfig.currency!];
     }
 
+    @XProvide('queriesPreviewInfo')
```
mnavarroespinosa, line 68: Should we watch this property? What would happen if `queriesPreview` is not defined initially in the `snippetConfig` and then is added programmatically through the `InterfaceX.setSnippetConfig` callback?

mnavarroespinosa (changes requested):
Not sure if you are already aware of it but there are some failing e2e tests.

src/components/pre-search/custom-query-preview.vue
```diff
@@ -0,0 +1,96 @@
...
+          </span>
+          <BaseEventButton
+            :events="getEvent(queryPreviewInfo.query)"
+            class="x-button x-button--ghost x-button--ghost-end"
+          >
+            {{ $t('queryPreview.viewResults') }}
+            <ChevronRightIcon />
+          </BaseEventButton>
+        </div>
+        <SlidingPanel :resetOnContentChange="false" :buttonClass="'x-button--ghost x-padding--00'">
```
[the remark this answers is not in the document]
ajperezbau: Yep! Both carousels should behave on the same way.

src/App.vue
ajperezbau: That shouldn't be needed because `setSnippetConfig` adds the properties reactively and the `XProvide` decorator keeps the queries previews also reactive.
I have just tested it, and it is working fine.

package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@empathyco/x-adapter": "^8.0.0-alpha.6",
     "@empathyco/x-adapter-platform": "^1.0.0-alpha.32",
-    "@empathyco/x-components": "3.0.0-alpha.162",
+    "@empathyco/x-components": "3.0.0-alpha.169",
```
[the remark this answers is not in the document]
ajperezbau: Yep, I have just updated all our internal dependencies, although I didn't update some packages to the last version because the archetype has to be updated in other tasks.
Feel free to check if I missed something 👌

src/components/pre-search/custom-query-preview.vue
```diff
@@ -0,0 +1,109 @@
...
+      SlidingPanel,
+      Result,
+      BaseEventButton,
+      ChevronLeftIcon,
+      ChevronRightIcon
+    }
+  })
+  export default class CustomQueryPreview extends Vue {
+    @XInject('queriesPreviewInfo')
+    public queriesPreviewInfo: QueryPreviewInfo[] | undefined;
```
tiborux, line 90:
```suggestion
    public queriesPreviewInfo!: QueryPreviewInfo[];
```

src/components/pre-search/custom-query-preview.vue
```diff
@@ -0,0 +1,109 @@
...
+
+    getEvent(query: string): any {
```
tiborux, line 92:
```suggestion
    getEvent(query: string): Partial<XEventsTypes> {
```

tiborux (changes requested):
gj

src/components/pre-search/custom-query-preview.vue
```diff
@@ -0,0 +1,110 @@
...
+            ? 'x-list--gap-01 x-margin--bottom-08'
+            : 'x-list--gap-05 x-margin--bottom-11'
+        "
+      >
+        <div :class="{ 'x-padding--left-05 x-padding--right-05': $x.device === 'mobile' }">
+          <h1 class="x-title1">{{ title }}</h1>
+          <div class="x-list x-list--horizontal x-list--align-center">
+            <span
+              class="x-font-weight--bold x-list__item--expand"
+              :class="$x.device === 'mobile' ? '' : 'x-title3 '"
```
mnavarroespinosa, line 22:
```suggestion
              :class="{ 'x-title3': $x.device === 'mobile' }"
```

mnavarroespinosa (approved):
Good job! It looks fantastic!

herrardo (approved):

PR #172, opened 2022-09-14
feat: showOnlyAfterOffset in next queries list

CachedaCodes:
EX-6982

## Motivation and context
Use the new prop for `NextQueriesList` `showOnlyAfterOffset`, that prevents next queries to be added when there's no more results to show after them.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Test with a query that returns fewer results than the offset (`hats`) and check the behavior of the prop.


src/components/search/results/results.vue
```diff
@@ -2,7 +2,12 @@
   <ResultsList v-if="$x.totalResults" v-infinite-scroll:main-scroll>
     <BannersList>
       <PromotedsList>
-        <NextQueriesList :offset="24" :frequency="48" :maxNextQueriesPerGroup="1">
+        <NextQueriesList
+          :offset="24"
+          :frequency="48"
+          :maxNextQueriesPerGroup="1"
+          :showOnlyAfterOffset="$x.partialResults.length === 0"
```
ajperezbau, line 9: I think that this condition should be inverted, because what we want to achieve is to show the NQ Previews when the results number is lesser than the offset:

```suggestion
          :showOnlyAfterOffset="$x.partialResults.length > 0"
```

ajperezbau (changes requested):

src/components/search/results/results.vue
mnavarroespinosa: I agree with @ajperezbau, I would invert the condition too

mnavarroespinosa (changes requested):

mnavarroespinosa (approved):
LGTM!

ajperezbau (approved):
Nice 👌

mnavarroespinosa (approved):

PR #177, opened 2022-09-20
Spike query preview in empathize

tajespasarela:
Test query preview in empathize


herrardo (commented):
How do we save this to use it in the future if we need it?

PR #181, opened 2022-09-30
fix(webpack): output webpack related files to `webpack-generated` virtual folder

diegopf:
EX-7221

This PR aims to solve the issue we currently have to debug `*.vue` files. In the devtools is tough to find what you want to debug through `cmd + p` since it lists several entries for the same file. This is due to webpack handling of this type of files. The vue-cli and vue repositories contain several issues related to the same problem (see for example: https://github.com/vuejs/vue-cli/issues/2897, https://github.com/vuejs/vue-cli/issues/2978 or https://github.com/vuejs/vue/issues/11023)

I have found the easiest solution for us is described on this [blog post](https://www.mistergoodcat.com/post/the-joy-that-is-source-maps-with-vuejs-and-typescript).

You can test this by comparing the content on the devtools:

Before
image

After
image

You can see now we have different virtual folders and when trying to open a file though `cmd + p` you should get it on the first entry (which is the one from `sources://`

After this PR is merged, the same can be done for the `x` monorepo


tajespasarela:
Really really cool!! 👏🏻 👏🏻 👏🏻  well done!

Maybe it is better to move to the `@empathyco/x-archetype-utils` and export the function to keep it maintained between the different setups?

diegopf:
> Really really cool!! 👏🏻 👏🏻 👏🏻 well done!
> 
> Maybe it is better to move to the `@empathyco/x-archetype-utils` and export the function to keep it maintained between the different setups?

I like how you think! I've created https://github.com/empathyco/x/pull/760 to export the whole webpack configuration from `x-archetype-utils`.

annacv:
Hi! I'm not sure if I'm doing something wrong, but in my local environment I still see all that App.vue .
This is how it looks like:
Captura de pantalla 2022-10-07 a les 9 09 51

I have checked PR preview link, and it's different 🤯 
Captura de pantalla 2022-10-07 a les 9 10 41

So it's probably something I'm missing. I have deleted `node_modules` and run `npm ci` again in case it fixed something, but it didn't work

package.json
```diff
@@ -20,6 +20,7 @@
     "lint": "eslint src tests build --ext .ts,.tsx,.vue,.js,.jsx",
     "lint:fix": "npm run lint -- --fix",
     "install:local": "npm install file:../x/packages/x-components/empathyco-x-components-3.0.0-alpha.178.tgz --no-save",
+    "install:utils:local": "npm install file:../x/packages/x-archetype-utils/empathyco-x-archetype-utils-0.1.0-alpha.2.tgz --no-save",
```
tiborux, line 23: `"@empathyco/x-archetype-utils": "^0.1.0-alpha.4"`

tiborux (changes requested):
gj

package.json
alvarodE: Done.

alvarodE (dismissed):

mnavarroespinosa (approved):
Good job!

tiborux (approved):

PR #185, opened 2022-10-11
feat: add styles to partial results

tiborux:
I know that some colors are not matching, cris is working on that. Right now the colors that we set in tailwind are not the same that the ones that the design has, so for example, the x-text is using neutral-10, instead of overwriting the token in the archetype what we have to do is set that value in tailwind.

What we want to achieve is that we should not override things in the archetype, everything has to be set as default in tailwind.

EX-4515

### DIFF
--- a/src/components/search/results/mobile-partial-results.vue
+++ b/src/components/search/results/mobile-partial-results.vue
@@ -18,7 +18,7 @@
         </span>
         <PartialQueryButton
           :query="partialResult.query"
-          class="x-title3 x-font-weight--bold x-button--ghost x-margin--left-auto"
+          class="x-text x-font-weight--bold x-button--ghost x-margin--left-auto"
         >
           {{ $t('partialResults.viewResults') }}
           <ChevronRightIcon />
@@ -62,7 +62,7 @@
       SlidingPanel
     }
   })
-  export default class PartialResults extends Vue {}
+  export default class MobilePartialResults extends Vue {}
 </script>
 
 <style lang="scss">
@@ -70,10 +70,6 @@
     .x-partial-results {
       &__result {
         width: calc(38vw - var(--x-size-gap-list-05));
-      }
-
-      .x-sliding-panel__scroll {
-        padding-left: var(--x-size-padding-grid);
       }
     }
   }
--- a/src/components/search/results/mobile-partial-results.vue
+++ b/src/components/search/results/mobile-partial-results.vue
@@ -29,7 +29,14 @@
         :buttonClass="'x-button--ghost x-padding--00'"
         class="x-sliding-panel--show-buttons-on-hover"
       >
-        <ResultsList :results="partialResult.results" />
+        <div class="x-list x-list--gap-05 x-padding--left-05">
+          <Result
+            v-for="result in partialResult.results"
+            :key="result.id"
+            :result="result"
+            class="x-partial-results__result"
+          />
+        </div>
       </SlidingPanel>
     </template>
   </PartialResultsList>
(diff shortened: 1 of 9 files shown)


src/design-system/tokens.scss
```diff
@@ -32,16 +32,14 @@
   --x-size-min-width-grid-item: var(--x-size-base-20);
   --x-size-gap-grid: var(--x-size-base-07) var(--x-size-base-05);
   --x-size-padding-grid: var(--x-size-base-06) 0;
+  .x-partial-result {
```
tiborux, line 35: this is not right, but if i change it without affecting just the partial-result, other grid will be affected.

src/components/search/results/mobile-partial-results.vue
```diff
@@ -0,0 +1,73 @@
...
+    components: {
+      ChevronRightIcon,
+      Result: ResultComponent,
+      PartialResultsList,
+      PartialQueryButton,
+      ResultsList,
+      SlidingPanel
+    }
+  })
+  export default class PartialResults extends Vue {}
```
CachedaCodes, line 65:
```suggestion
  export default class MobilePartialResults extends Vue {}
```

src/components/search/results/mobile-partial-results.vue
```diff
@@ -0,0 +1,73 @@
+<template>
```
CachedaCodes, line 1: Is it worth it to have separate files for desktop and mobile? The only major difference is the grid instead of the sliding panel (the other ones are two classes and a prop).
I like the implementation, but having them both in the same file is not too convoluted and is convenient if we have to make changes (we won't have to make them twice).

src/components/search/results/mobile-partial-results.vue
```diff
@@ -0,0 +1,80 @@
...
+<style lang="scss">
+  .x-mobile {
+    .x-partial-results {
+      &__result {
+        width: calc(38vw - var(--x-size-gap-list-05));
+      }
+
+      .x-sliding-panel__scroll {
+        padding-left: var(--x-size-padding-grid);
+      }
```
CachedaCodes, line 77: This rule is not doing anything.

src/design-system/tokens.scss
CachedaCodes: Should the partials grid follow the same directives as the other grids (talking about the design here)?

CachedaCodes (changes requested):

src/components/search/results/partial-results.vue
```diff
@@ -1,32 +1,55 @@
...
+      ChevronRightIcon,
       Result: ResultComponent,
       PartialResultsList,
       PartialQueryButton
     }
   })
-  export default class PartialResults extends Vue {}
+  export default class PartialResults extends Vue {
+    @Prop({ default: 4 })
+    protected maxItemsToRender!: number;
```
diegopf, line 53: why do we need this prop? as far as I can see it is not even used.

src/components/search/results/mobile-partial-results.vue
diegopf: I'm not sure about that, I'd rather have two separate files than one and having to check what is applying on mobile or desktop. I think the agreement is to have as less logic as possible in the archetype and having two different files is better for that.

src/design-system/tokens.scss
diegopf: yeah, that would be nice. Maybe we can talk with cris about it

diegopf (changes requested):

src/components/search/results/mobile-partial-results.vue
tiborux: yep, @tajespasarela  said that is better to have 2 files instead of having one and play with props, slots and other things

src/components/search/results/partial-results.vue
tiborux: Finally I remove it the prop, this should be done in procut

CachedaCodes (approved):

diegopf (dismissed):

diegopf (approved):

PR #187, opened 2022-10-12
feat: enable `devtools` in Empathy environments

JaviIG:
- Enables devtools as long as the `__enableVueDevtools__` variable is `true`. 
- Sets the `__enableVueDevtools__` in Empathy environments.
- Fixes compilation in Windows devices.

### DIFF
--- a/build/instrumentation.build.js
+++ b/build/instrumentation.build.js
@@ -104,7 +104,10 @@ export function createConfig({
             'inject',
             (varname, id) =>
               // eslint-disable-next-line max-len
-              `import {createInjector} from 'vue-runtime-helpers';const injector=createInjector({});injector('${id}',{source:${varname}})`
+              `import {createInjector} from 'vue-runtime-helpers';const injector=createInjector({});injector('${id.replace(
+                /\\/g,
+                ''
+              )}',{source:${varname}})`
           ]
         })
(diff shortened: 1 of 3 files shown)


mavmaf (approved):

build/instrumentation.build.js
```diff
@@ -104,7 +104,10 @@ export function createConfig({
             'inject',
             (varname, id) =>
               // eslint-disable-next-line max-len
-              `import {createInjector} from 'vue-runtime-helpers';const injector=createInjector({});injector('${id}',{source:${varname}})`
+              `import {createInjector} from 'vue-runtime-helpers';const injector=createInjector({});injector('${id.replace(
+                /\\/g,
+                ''
```
tajespasarela, line 109: liam-neeson-why (1)

tajespasarela (changes requested):

build/instrumentation.build.js
javieri-empathy: Because the `id` is a path, an in windows they use backslash and break things due to scaping. It is done the same way in X. I can rollback the change or document it if you prefer

mavmaf (approved):

tajespasarela (approved):

PR #200, opened 2022-10-24
feat: update archetype to not use facets config

ajperezbau:
Also use facets model name in the custom facets component slots.

[EX-7259](https://searchbroker.atlassian.net/browse/EX-7259)

### DIFF
--- a/src/components/search/facets/custom-facets.vue
+++ b/src/components/search/facets/custom-facets.vue
@@ -32,7 +32,7 @@
       </CustomHeaderTogglePanel>
     </template>
 
-    <template #category-paths="{ facet, selectedFilters }">
+    <template #hierarchical-facet="{ facet, selectedFilters }">
       <CustomHeaderTogglePanel
         :data-test="facet.label"
         class="x-border-width--bottom-01 x-border-width--00 x-border-color--neutral-95"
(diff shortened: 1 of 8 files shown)


src/components/search/facets/custom-facets.vue
```diff
@@ -32,7 +32,7 @@
       </CustomHeaderTogglePanel>
     </template>
 
-    <template #category-paths="{ facet, selectedFilters }">
+    <template #hierarchical-facet="{ facet, selectedFilters }">
```
tiborux, line 35: This is not right, the slot uses the id of the facet:

image

So you have to rollback these names

tiborux:
Update `x-platform-adapter` version

annacv (commented):
Hey, @tiborux I'm not sure about this, if we used the `modelName`, wouldn't we avoid repeating the template for facets that share that model??? Maybe I'm missing something... are we constrained to use the `id` for some reason?

tiborux:
> Hey, @tiborux I'm not sure about this, if we used the `modelName`, wouldn't we avoid repeating the template for facets that share that model??? Maybe I'm missing something... are we constrained to use the `id` for some reason?

Yes, you are right, I didn't check the component deeply, I saw that we use the id for the slots, but we also use the model name for the slots,

alvarodE (approved):

alvarodE (approved):

tiborux (approved):

PR #206, opened 2022-11-11
feat: replace x-button with XDS x-button implementation

CachedaCodes:
EX-5908

## Motivation and context
Add `x-button` implementation from the XDS

## Type of change

- [X] `Main`
- [ ] Other. Specify:

### DIFF
--- a/src/components/search/results/mobile-partial-results.vue
+++ b/src/components/search/results/mobile-partial-results.vue
@@ -18,7 +18,7 @@
         </span>
         <PartialQueryButton
           :query="partialResult.query"
-          class="x-text x-font-weight--bold x-button--ghost x-margin--left-auto"
+          class="x-button-lead x-button-ghost x-margin--left-auto"
         >
           {{ $t('partialResults.viewResults') }}
           <ChevronRightIcon />
@@ -26,7 +26,7 @@
       </div>
       <SlidingPanel
         :resetOnContentChange="false"
-        :buttonClass="'x-button--ghost x-padding--00'"
+        :buttonClass="'x-button-ghost x-padding--00'"
         class="x-sliding-panel--show-buttons-on-hover"
       >
         <div class="x-list x-list--gap-05 x-padding--left-05">
--- a/src/components/search/results/mobile-partial-results.vue
+++ b/src/components/search/results/mobile-partial-results.vue
@@ -26,7 +26,7 @@
       </div>
       <SlidingPanel
         :resetOnContentChange="false"
-        :buttonClass="'x-button-ghost x-padding--00'"
+        buttonClass="x-button-lead x-button-circle x-button-ghost x-padding--00"
         class="x-sliding-panel--show-buttons-on-hover"
       >
         <div class="x-list x-list--gap-05 x-padding--left-05">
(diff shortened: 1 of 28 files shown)


herrardo (changes requested):
I am comparing this PR Preview with what we have deployed in test. I see some differences.

1. There is different margin between suggestions, i guess because history queries have a button inside
image

2. Have we change the font family? I see different behavior
image

3. The difference between the magnifying glass button on the search input is on purpose?

image

CachedaCodes:
@herrardo 

> I am comparing this PR Preview with what we have deployed in test. I see some differences.
> 
> 1. There is different margin between suggestions, i guess because history queries have a button inside

Yes, a quick fix was made in the PR, and we'll deal with it when we implement the suggestion group

> 2. Have we change the font family? I see different behavior

The one here is Montserrat, which I think is correct, the one in x.test is default sans-serif for some reason. Anyway, it should be Montserrat.

> 3. The difference between the magnifying glass button on the search input is on purpose?

Fixed, thanks!

herrardo:
> @herrardo
> 
> > I am comparing this PR Preview with what we have deployed in test. I see some differences.
> > 
> > 1. There is different margin between suggestions, i guess because history queries have a button inside
> 
> Yes, a quick fix was made in the PR, and we'll deal with it when we implement the suggestion group
> 
> > 2. Have we change the font family? I see different behavior
> 
> The one here is Montserrat, which I think is correct, the one in x.test is default sans-serif for some reason. Anyway, it should be Montserrat.
> 
> > 3. The difference between the magnifying glass button on the search input is on purpose?
> 
> Fixed, thanks!

I think the font-family is ok in test environment, we are using **Inter** font, but ask @cris-crash to be sure.

CachedaCodes:
There are known issues with the XDS button implementation that I'm currently discussing with design. I will deal with them afterwards.

Now that the `x-button` has hover effects, the size created inside of the button by its padding is noticeable in the `x-button-ghost`. In the past we used the padding of the button to align it with other elements when needed (removing it, for example), doing that now creates this effect

image

Clear button in the searchbox is part of an input group that forces some styles (same bg color for example). The `x-button` class adds an on hover effect that changes colors, resulting in this effect when hovering the clear button.

image

Sometimes, the new hover effects and its interaction with the `x-button-ghost` produce weird results depending on the background the button is or the elements inside the icon.

image

image

The button size changing makes it so elements previously aligned now are not.

image

Some combinations of bg color and font are not possible with the DS or clash with it

image

image

src/components/search/results/mobile-partial-results.vue
```diff
@@ -18,15 +18,15 @@
         </span>
         <PartialQueryButton
           :query="partialResult.query"
-          class="x-text x-font-weight--bold x-button--ghost x-margin--left-auto"
+          class="x-button-lead x-button-ghost x-margin--left-auto"
```
tiborux, line 21: Removing the x-text the font-size changes, I supposed that this is right

tiborux (changes requested):
gj

tiborux:
You missed the next-query:

image

mnavarroespinosa (changes requested):
Good job!

CachedaCodes:
> You missed the next-query:
> 
> image

This one is technically a suggestion, which will be its own component, so it shouldn't be changed.

tiborux (approved):
gj

mnavarroespinosa (approved):

tiborux (approved):

PR #210, opened 2022-11-17
fix: fix styles for the selected filters reverting previous changes

CachedaCodes:
EX-7431

## Motivation and context
Selected filters are tags, so they should be changed to buttons (they'll be their own component). The changes in the previous PR should be reverted.
Also, removing the custom paddings for the clear filters, the mismatch between tags and buttons sizes should be fixed with the tag component.

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)

- [X] `Main`
- [ ] Other. Specify:

### DIFF
--- a/src/components/search/facets/selected-filters.vue
+++ b/src/components/search/facets/selected-filters.vue
@@ -23,8 +23,8 @@
             :filter="filter"
           >
             <template #label="{ filter }">
-              <span>{{ filter.label }}</span>
-              <CrossTinyIcon />
+              <span class="x-text x-text--bold x-font-color--accent">{{ filter.label }}</span>
+              <CrossTinyIcon class="x-font-color--accent" />
             </template>
           </SimpleFilter>
         </template>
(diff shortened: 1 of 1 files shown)


tiborux (approved):
gj

src/components/search/facets/selected-filters.vue
```diff
@@ -23,25 +23,24 @@
             :filter="filter"
           >
             <template #label="{ filter }">
-              <span>{{ filter.label }}</span>
-              <CrossTinyIcon />
+              <span class="x-text x-text--bold x-font-color--accent">{{ filter.label }}</span>
```
tiborux, line 26: Clear selected filters doesnt have the same height as the others

image

tiborux: Accepted by design, go ahead

tiborux (dismissed):

src/components/search/facets/selected-filters.vue
```diff
@@ -55,13 +54,7 @@
...
       v-slot="{ selectedFilters }"
       data-test="clear-filters-toolbar"
-      class="
-        x-button-sm x-button-lead x-button-outlined
-        x-border-radius--20
-        x-padding--05 x-padding--top-03 x-padding--bottom-03
-        x-margin--left-03
-        x-list__item--flex-none
-      "
+      class="x-button-sm x-button-lead x-button-outlined x-border-radius--20"
```
mnavarroespinosa, line 57: Removing `x-list__item--flex-none` may cause the `Clear filters` button to wrap the text when there are many selected items. But it is maybe a corner case, I'm fine if we don't add the class.

image

mnavarroespinosa (approved):
good job!

tiborux (approved):

PR #213, opened 2022-11-18
feat: Display Spellckeck above results toolbar

alvarodE:
[EX-7243](https://searchbroker.atlassian.net/browse/EX-7243)

With this change we are losing the origin when clicking on the spellcheck button:
- Before:
Screenshot 2022-11-18 at 09 06 09
- After:
Screenshot 2022-11-18 at 09 06 21

I think that origin doesn't represent anything relevant in the case of a spellcheck, but happy to read different opinions :)

### DIFF
--- a/src/components/desktop/desktop.vue
+++ b/src/components/desktop/desktop.vue
@@ -38,10 +38,10 @@
     </template>
 
     <template #sub-header v-if="!$x.redirections.length">
-      <LocationProvider location="no_results">
+      <LocationProvider location="results">
         <SpellcheckMessage class="x-margin--bottom-05" data-test="spellcheck-message" />
-        <NoResultsMessage class="x-margin--bottom-05" data-test="no-results-message" />
       </LocationProvider>
+      <NoResultsMessage class="x-margin--bottom-05" data-test="no-results-message" />
       <DesktopToolbar />
     </template>
 
--- a/src/components/desktop/desktop.vue
+++ b/src/components/desktop/desktop.vue
@@ -38,8 +38,10 @@
     </template>
 
     <template #sub-header v-if="!$x.redirections.length">
-      <SpellcheckMessage class="x-margin--bottom-05" data-test="spellcheck-message" />
-      <NoResultsMessage class="x-margin--bottom-05" data-test="no-results-message" />
+      <LocationProvider location="no_results">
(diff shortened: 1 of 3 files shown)


src/components/desktop/desktop.vue
```diff
@@ -37,7 +37,9 @@
       </div>
     </template>
 
-    <template #sub-header>
+    <template #sub-header v-if="!$x.redirections.length">
+      <SpellcheckMessage class="x-margin--bottom-05" data-test="spellcheck-message" />
+      <NoResultsMessage class="x-margin--bottom-05" data-test="no-results-message" />
```
mnavarroespinosa, line 42: What would you think of wrapping this inside a `LocationProvider` to keep sending the `results` location? (and the same for mobile).

I'm not sure either if it is really relevant in this case, but I think I would add it so the archetype is more complete.

src/components/mobile/mobile.vue
```diff
@@ -33,8 +33,18 @@
       </div>
     </template>
 
-    <template #toolbar v-if="$x.query.search">
-      <MobileToolbar class="x-padding--left-05 x-padding--bottom-05 x-padding--right-03" />
+    <template #toolbar v-if="$x.query.search && !$x.redirections.length">
+      <div>
```
mnavarroespinosa, line 37: The `div` is not taking the full width

```suggestion
      <div class="x-list__item--expand">
```

image

mnavarroespinosa (changes requested):

src/components/desktop/desktop.vue
herrardo: I would say that the correct Location would be `no_results`  as described in the types:

https://github.com/empathyco/x/blob/9c235bc9457cc959fa2a386b8dce114ff3bf6df7/packages/x-components/src/types/origin.ts#L62

And correct origin would be `spellcheck`:

https://github.com/empathyco/x/blob/9c235bc9457cc959fa2a386b8dce114ff3bf6df7/packages/x-components/src/types/origin.ts#L32

herrardo (changes requested):

mnavarroespinosa (approved):
Good job!

CachedaCodes (approved):

PR #226, opened 2022-12-30
fix: no-results are now correctly displayed

javieri-empathy:
EX-7717

The rendering condition was wrong. Now the no-results reccommendations are correctly displayed:

Before:
image

After:
image


mavmaf (approved):
Thanks for adding the step to the test :)
Closing EX-7534 as duplicated

CachedaCodes (approved):

PR #232, opened 2023-01-05
test: re-design scroll test

mavmaf:
EX-7762

## Motivation and context
Scroll test was either failing or throwing a false positive when comparing two lists of results using different kind of sorting. As testing the sorting isn't the purpose of the test but just a way to check the scrolling behaviour, the test has been redesign in order not to use the sorting options. With these changes, scrolled position persistence is tested by reloading the page; and scrolled position reset is checked both using scroll to top, and launching a new search by selecting the related tags.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## Checklist:

- [x] My code follows the **[style guidelines](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md#style-guides)** of this project.
- [x] My changes generate **no new warnings**.
- [x] I have added **tests** that prove my fix is effective or that my feature works.
- [x] New and existing **unit tests pass locally** with my changes.


tests/e2e/cucumber/scroll.spec.ts
```diff
@@ -1,16 +1,29 @@
...
-    .scrollIntoView({ offset: { top: -5, left: 0 } })
-    .should('be.visible');
+  cy.get(`[data-scroll=${resultId}]`).eq(0).scrollIntoView();
 });
 
 Then('url is updated with result {string}', (resultId: string) => {
   cy.url().should('contain', `scroll=${resultId}`);
 });
 
+When('the page is reloaded', () => {
```
ajperezbau, line 11: Should this `When` be placed in the `common-steps` file?

tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -29,7 +29,11 @@ Given('a results API with known results', () => {
     'interceptedResults'
   );
 });
-
+Given('a results API with known results', () => {
```
ajperezbau, line 32: This seems to be duplicated

tests/e2e/cucumber/scroll.spec.ts
```diff
@@ -1,16 +1,29 @@
...
+
+When('scroll to top is clicked', () => {
+  cy.getByDataTest('scroll-to-top').click();
+});
+
 Then('scroll position is at top', () => {
   cy.url().should('not.contain', 'scroll');
 });
+
+Then('{string} pdp link is clicked', (resultId: string) => {
```
ajperezbau, line 27: Should we remove this?
It seems that it is not being used.

ajperezbau (changes requested):
Good job! 👌

tests/e2e/cucumber/mocked-responses.spec.ts
LuisMartinez15: +1

tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -29,7 +29,11 @@ Given('a results API with known results', () => {
...
+  cy.intercept(searchEndpoint, { fixture: 'search-response-with-known-results' }).as(
+    'interceptedResults'
+  );
+});
```
LuisMartinez15, line 36: **issue:** this step is duplicated

```suggestion
```

tests/e2e/cucumber/scroll.spec.ts
LuisMartinez15: If it is only used here I don't see a reason to move it to the `common-steps` file.

tests/e2e/cucumber/scroll.spec.ts
LuisMartinez15: agree

LuisMartinez15 (changes requested):
**praise:** good jobı

tests/e2e/cucumber/scroll.spec.ts
mavmaf: Although the action is generalist it's only being applied in this test and moving it could reduce step findability (translation: if we move it to `common-steps` it might take more time to find it than where it is now)

LuisMartinez15 (approved):

ajperezbau (approved):

PR #240, opened 2023-01-16
feat(history-queries): hide history queries when there are none with results

ajperezbau:
I have moved the condition to show the predictive layer from the `Empathize` component to the `div` containing the suggestions because the `Empathize` already has some logic to hide the predictive layer if it doesn't have content.

This fixes the behaviour where the `Empathize` is not displayed after clearing a query with no results and a history queries test that was failing occasionally because the `Empathize` was closed and Cypress couldn't get the history queries elements.

EX-7720


annacv (approved):

annacv:
I tried a fix on this test `3. Clear a history query` test on `predictive-components.feature` but it didn't work :(

annacv (dismissed):

annacv (approved):
🎉

LuisMartinez15 (approved):

PR #246, opened 2023-01-24
feat(design-system): use sliding panel from XDS

diegopf:
EX-7832

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@empathyco/x-adapter": "^8.0.0-alpha.17",
     "@empathyco/x-adapter-platform": "^1.0.0-alpha.58",
     "@empathyco/x-archetype-utils": "^0.1.0-alpha.13",
-    "@empathyco/x-components": "^3.0.0-alpha.286",
+    "@empathyco/x-components": "^3.0.0-alpha.295",
     "@empathyco/x-deep-merge": "^1.3.0-alpha.26",
     "@empathyco/x-types": "^10.0.0-alpha.50",
     "@empathyco/x-utils": "^1.0.0-alpha.13",
@@ -43,7 +43,7 @@
   "devDependencies": {
     "@cypress/webpack-preprocessor": "~5.11.1",
     "@empathyco/eslint-plugin-x": "^2.0.0-alpha.21",
-    "@empathyco/x-tailwindcss": "^0.2.0-alpha.44",
+    "@empathyco/x-tailwindcss": "^0.2.0-alpha.54",
     "@empathyco/x-translations": "^1.1.0-alpha.30",
     "@rollup/plugin-commonjs": "~21.0.1",
     "@rollup/plugin-json": "~4.1.0",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "devDependencies": {
     "@cypress/webpack-preprocessor": "~5.11.1",
     "@empathyco/eslint-plugin-x": "^2.0.0-alpha.21",
-    "@empathyco/x-tailwindcss": "^0.2.0-alpha.54",
+    "@empathyco/x-tailwindcss": "^0.2.0-alpha.56",
     "@empathyco/x-translations": "^1.1.0-alpha.30",
     "@rollup/plugin-commonjs": "~21.0.1",
     "@rollup/plugin-json": "~4.1.0",
--- a/src/components/mobile/mobile-next-query-preview.vue
+++ b/src/components/mobile/mobile-next-query-preview.vue
@@ -48,7 +48,7 @@
       :showButtons="false"
       :resetOnContentChange="false"
     >
-      <div class="x-list x-list--gap-05">
+      <div class="x-flex x-flex-row x-gap-16">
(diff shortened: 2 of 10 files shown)


src/components/predictive-layer/sliding-recommendations.vue
```diff
@@ -9,9 +9,10 @@
       </h1>
```
ajperezbau, line 9: I think that the `style` block of this file is no longer needed.

src/components/pre-search/custom-query-preview.vue
```diff
@@ -36,13 +36,14 @@
...
-          class="x-sliding-panel--show-buttons-on-hover"
+          scrollContainerClass="x-sliding-panel-fade-sm"
+          class="x-sliding-panel-show-buttons-on-hover"
         >
           <template #sliding-panel-left-button>
             <ChevronLeftIcon class="x-icon--l" />
           </template>
           <ul
-            class="x-list x-list--gap-05"
+            class="x-flex x-flex-row x-gap-16 x-sliding-panel-fade-sm"
```
ajperezbau, line 46:
```suggestion
            class="x-flex x-gap-16"
```

package.json
```diff
@@ -28,7 +28,7 @@
     "@empathyco/x-adapter": "^8.0.0-alpha.17",
     "@empathyco/x-adapter-platform": "^1.0.0-alpha.58",
```
ajperezbau, line 29: Now that I see the fade better, it seems that it is making all the content of the sliding panel brighter.
It was like that before too, but, should we modify it in another task to only affect the borders?

https://user-images.githubusercontent.com/75546736/214264941-7f2529b8-d49f-44e5-b843-4cf7f050e949.mov

ajperezbau (changes requested):
It seems a bit strange to me having to set the `sliding-panel-fade` classes with the `scrollContainerClass`, shouldn't it go in the `sliding-panel` itself? 🤔 

Good job! 👌

src/components/search/results/mobile-partial-results.vue
```diff
@@ -29,7 +29,7 @@
       <SlidingPanel
         :resetOnContentChange="false"
         buttonClass="x-button-lead x-button-circle x-button-ghost x-padding--00"
-        class="x-sliding-panel--show-buttons-on-hover"
+        class="x-sliding-panel-show-buttons-on-hover"
       >
         <div class="x-list x-list--gap-05 x-padding--left-05">
```
LuisMartinez15, line 34: **issue:** `partial-results` are broken in mobile, they're displayed in a column instead of a row.

```suggestion
        <div class="x-list x-list--horizontal x-list--gap-05 x-padding--left-05">
```

LuisMartinez15 (changes requested):
**praise:** good job!

src/components/mobile/mobile-next-query-preview.vue
```diff
@@ -48,7 +48,7 @@
       :showButtons="false"
```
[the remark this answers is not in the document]
diegopf: You mean the `padding-left: var(--x-size-padding-grid);` ? I don't think we can remove that, the XDS sliding panel doesn't care about that padding.

src/components/search/related-tags.vue
```diff
@@ -1,7 +1,8 @@
 <template>
```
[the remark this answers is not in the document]
diegopf: This needs to be fixed on XDS, see https://github.com/empathyco/x/pull/1034

package.json
diegopf: IMO I would create a different task to address this.

src/components/mobile/mobile-next-query-preview.vue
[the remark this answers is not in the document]
ajperezbau: Oh, I see that we are still using the `x-sliding-panel__scroll` in the component.
Nevermind, I would add that padding directly using the `scrollContainerClass="x-pl-16"` instead of with an internal selector.

The padding plugin needs to be enabled.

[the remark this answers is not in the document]
ajperezbau: Well, actually we can use `x-px-16` to apply it to the right too.

src/components/pre-search/custom-query-preview.vue
```diff
@@ -36,13 +36,14 @@
           :showButtons="$x.device !== 'mobile'"
           :resetOnContentChange="false"
           buttonClass="x-button-lead x-button-circle x-button-ghost x-padding--00"
-          class="x-sliding-panel--show-buttons-on-hover"
+          scrollContainerClass="desktop:x-sliding-panel-fade-sm"
```
ajperezbau, line 39: I see that this is the only sliding panel using an `sm` fade, is that ok?

ajperezbau (commented):
Nice 👌

ajperezbau (approved):

ajperezbau (approved):

cris-crash (commented):
good job!

PR #247, opened 2023-01-24
feat: Implement XDS product image in archetype

annacv:
EX-7837

- implement XDS class, 
- update packages, 
- add `backgroundColor` plugin in tailwind to support `x-picture-overlay` variant

[Sliding panel PR](https://github.com/empathyco/x-archetype/pull/246) should be merged first and then update this branch with that changes. Currently, tagging tests are failing as the row gets unstyled and then the list of products can't be scrolled as expected.

Also, the [XDS use class instead of img tag PR](https://github.com/empathyco/x/pull/1035) should be merged first and then update the package version.


mnavarroespinosa (dismissed):
Good job!

diegopf (approved):

mnavarroespinosa (approved):

PR #248, opened 2023-01-24
fix: Add husky to run gitHooks

annacv:
Recently `gitHooks` were failing to find `yorkie` and the commits weren't able to be performed. 
This is because that dependency was removed from `@vue/cli-plugin-typescript`.
To address it we have added `husky`, as it has more maintenance.

Developed with @mnavarroespinosa

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "cypress-cucumber-preprocessor": "~4.3.1",
     "eslint": "~8.30.0",
     "http-server": "^14.1.0",
-    "husky": "^8.0.3",
+    "husky": "~8.0.3",
     "lint-staged": "~11.2.6",
     "postcss": "~8.3.11",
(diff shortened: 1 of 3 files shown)


mnavarroespinosa (approved):

package.json
```diff
@@ -66,6 +67,7 @@
     "cypress-cucumber-preprocessor": "~4.3.1",
     "eslint": "~8.30.0",
     "http-server": "^14.1.0",
+    "husky": "^8.0.3",
```
diegopf, line 70: I think we should use ~version and not ^version

diegopf (changes requested):
**praise**: nice PR!

diegopf (approved):

mnavarroespinosa (approved):

PR #260, opened 2023-02-10
feat:(design-system): Use `input-group` from `XDS`

tajespasarela:
EX-7943

### DIFF
--- a/src/components/search-box.vue
+++ b/src/components/search-box.vue
@@ -1,26 +1,27 @@
 <template>
-  <div class="x-input-group x-input-group--card x-border-color--lead">
+  <div class="x-input-group x-input-group-lead x-rounded-sm">
     <SearchInput
       :autofocus="false"
       :placeholder="$t('searchBox.placeholder')"
-      :class="$x.device === 'mobile' ? 'x-padding--left-05' : 'x-padding--left-08'"
+      :instant="false"
+      class="desktop:x-pl-40"
     />
 
-    <!-- TODO: Remove x-max-h when implementing XDS input-group -->
     <ClearSearchInput
       v-if="$x.device === 'desktop' && $x.query.searchBox"
-      class="x-button-lead !x-button-sm x-button-ghost x-max-h-32"
+      class="x-input-group-button"
     >
       {{ $t('searchBox.clear') }}
     </ClearSearchInput>
 
-    <button
+    <SearchButton
       v-if="$x.device === 'desktop' || !$x.query.searchBox"
-      class="x-input-group__action x-button x-button-lead x-button-square"
+      class="x-input-group-button-primary"
     >
       <SearchIcon class="x-icon-lg" />
-    </button>
-    <ClearSearchInput v-else class="x-input-group__action x-button-lead x-button-square">
+    </SearchButton>
+
+    <ClearSearchInput v-else class="x-input-group-button-primary">
       <CrossTinyIcon class="x-icon-lg" />
     </ClearSearchInput>
   </div>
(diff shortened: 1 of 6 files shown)


src/components/search-box.vue
```diff
@@ -1,26 +1,27 @@
 <template>
-  <div class="x-input-group x-input-group--card x-border-color--lead">
+  <div class="x-input-group x-input-group-lead x-rounded-sm">
     <SearchInput
       :autofocus="false"
       :placeholder="$t('searchBox.placeholder')"
-      :class="$x.device === 'mobile' ? 'x-padding--left-05' : 'x-padding--left-08'"
+      :instant="false"
+      class="desktop:x-pl-40"
```
CachedaCodes, line 7: The input group's first child now has a margin of `16px`, so now the input is 16px more to the right than it should be in desktop (in mobile is ok).

```suggestion
      class="desktop:x-pl-24"
```

src/components/search-box.vue
```diff
@@ -1,26 +1,27 @@
 <template>
-  <div class="x-input-group x-input-group--card x-border-color--lead">
+  <div class="x-input-group x-input-group-lead x-rounded-sm">
     <SearchInput
```
CachedaCodes, line 3: Font size has changed from `16px` to inheriting `14px`, I think we should change it in the home template from `x-archetype-utils` so they're the same.

CachedaCodes (changes requested):
Nice 👍

src/components/search-box.vue
```diff
@@ -1,26 +1,27 @@
...
     <ClearSearchInput
       v-if="$x.device === 'desktop' && $x.query.searchBox"
-      class="x-button-lead !x-button-sm x-button-ghost x-max-h-32"
+      class="x-input-group-button"
     >
       {{ $t('searchBox.clear') }}
     </ClearSearchInput>
 
-    <button
+    <SearchButton
```
CachedaCodes, line 17: The search button is now bigger than the one in the `home-template` so the same as above with the font

src/components/search-box.vue
```diff
@@ -1,26 +1,27 @@
 <template>
-  <div class="x-input-group x-input-group--card x-border-color--lead">
+  <div class="x-input-group x-input-group-lead x-rounded-sm">
```
ajperezbau, line 2: I don't know if the designs are updated, but there the input group has the neutral 90 color.

```suggestion
  <div class="x-input-group x-rounded-sm">
```

ajperezbau (commented):
The empathize is rendering a double scroll on mobile, this will be solved when [this PR](https://github.com/empathyco/x-archetype/pull/262) is merged.

Should we wait till that to merge this one?

Looks good to me apart from the comments of @CachedaCodes 👌

src/components/search-box.vue
tajespasarela: the designs we have to match are these of the XDS: https://www.figma.com/file/9a1DMgqIHt5ZZBvS2waaOs/X-DS?node-id=4126%3A81277&t=mcdh9xFjIwKM0MlH-0

src/components/search-box.vue
tajespasarela: fixed

src/components/search-box.vue
tajespasarela: I don't see it bigger. It is 40px x 40px in both cases. Also the icon is 24px in both cases

src/components/search-box.vue
tajespasarela: done in this other PR: https://github.com/empathyco/x/pull/1079

ajperezbau (approved):

CachedaCodes (approved):

PR #262, opened 2023-02-15
feat(design-system): Add new layouts from XDS.

tajespasarela:
EX-7792

### DIFF
--- a/src/components/predictive-layer/predictive-layer.vue
+++ b/src/components/predictive-layer/predictive-layer.vue
@@ -1,6 +1,10 @@
 <template>
   <Empathize :animation="empathizeAnimation" class="x-bg-neutral-0 desktop:x-py-16 desktop:x-pl-16">
-    <component :is="$x.device === 'desktop' ? 'div' : 'BaseScroll'" class="x-h-full">
+    <component
+      :is="$x.device === 'desktop' ? 'div' : 'BaseScroll'"
+      v-if="showEmpathize || showIdentifierResults"
+      class="x-h-full"
+    >
       <div class="x-layout-item desktop:x-block">
         <BaseKeyboardNavigation
           class="x-flex x-flex-col desktop:x-grid desktop:x-grid-cols-12 desktop:x-items-start desktop:x-gap-24"
@@ -162,14 +166,11 @@
 
           <SlidingRecommendations
             v-if="$x.device === 'desktop' && !$x.query.searchBox"
-            class="desktop:x-col-span-7"
+            class="x-col-span-7"
           />
         </BaseKeyboardNavigation>
       </div>
-      <SlidingRecommendations
-        v-if="$x.device === 'mobile' && !$x.query.searchBox"
-        class="desktop:x-col-span-7"
-      />
+      <SlidingRecommendations v-if="$x.device === 'mobile' && !$x.query.searchBox" />
     </component>
   </Empathize>
 </template>
--- a/src/components/predictive-layer/sliding-recommendations.vue
+++ b/src/components/predictive-layer/sliding-recommendations.vue
@@ -1,7 +1,7 @@
 <template>
   <div v-if="$x.recommendations.length > 0" class="x-list x-list--gap-04">
     <h1
-      class="x-uppercase x-title4 x-title4-sm x-flex x-h-32 x-items-center x-pl-16 desktop:x-pl-0"
+      class="x-uppercase x-title4 x-title4-sm x-flex x-h-32 x-items-center x-pt-16 x-pl-24 desktop:x-pl-0 desktop:x-pt-0"
     >
       {{ $t('recommendations.title') }}
     </h1>


herrardo (changes requested):
1- Is a side effect of this PR that the margins have been reduced and now we have more products on BrandRecommendations/QueryPreview?

If so, we should restore them

image

2- If I close the Archetype and see the home again, I can't reopen it.

herrardo:
> 1- Is a side effect of this PR that the margins have been reduced and now we have more products on BrandRecommendations/QueryPreview?
> 
> If so, we should restore them
> 
> image
> 
> 2- If I close the Archetype and see the home again, I can't reopen it.

After talking to @cris-crash and @LiciCarvil  we think there was a misunderstanding, the max-width of the grid should be 1440.

CachedaCodes (changes requested):

tailwind.config.js
```diff
@@ -18,17 +19,43 @@ module.exports = {
   corePlugins: [
```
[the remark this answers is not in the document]
tajespasarela: I would remove it in the task of changing all the old `x-list-...` `x-row-...`

src/components/desktop/desktop.vue
```diff
@@ -1,94 +1,90 @@
 <template>
-  <FixedHeaderAndAsidesLayout :class="{ 'x-background--neutral-100': !!$x.query.searchBox }">
-    <template #header>
-      <div
-        class="x-list__item--expand x-row x-row--align-start x-row--gap-04 x-padding--top-06 x-padding--bottom-05"
-      >
-        <div class="x-row__item x-row__item--span-2 x-list x-padding--top-04">
+  <div
+    class="x-layout-container x-layout-max-width-md x-layout-min-margin-16 tablet:x-layout-min-margin-24 desktop:x-layout-min-margin-32 large:x-layout-max-width-lg large:x-layout-min-margin-48"
```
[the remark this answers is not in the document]
tajespasarela: where are we removing that breakpoint and why?

src/components/desktop/desktop.vue
```diff
@@ -1,94 +1,90 @@
...
+  >
+    <div class="x-layout-item">
+      <header class="x-align-start x-grid x-grid-cols-6 x-gap-12 x-pt-24 x-pb-16">
+        <div class="x-list x-pt-12">
```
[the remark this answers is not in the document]
tajespasarela: yes! good catch

src/components/desktop/desktop.vue
```diff
@@ -1,94 +1,90 @@
...
           <Logo />
         </div>
 
-        <div class="x-row__item x-row__item--span-8 x-list x-list--vertical x-list--gap-05">
+        <div class="x-col-span-4 x-flex x-flex-col x-gap-16">
```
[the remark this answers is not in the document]
tajespasarela: without flex and gap:
image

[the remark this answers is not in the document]
tajespasarela: I can replace it with margin to in the RTs if you prefer

tajespasarela:
> After talking to @cris-crash and @LiciCarvil we think there was a misunderstanding, the max-width of the grid should be 1440.

changed

tajespasarela:
> 2- If I close the Archetype and see the home again, I can't reopen it.

Fixed in other task/PR

herrardo (dismissed):

src/components/mobile/mobile.vue
```diff
@@ -1,162 +1,121 @@
 <template>
-  <SingleColumnLayout
-    :asideAnimation="animation"
-    :class="{ 'x-background--neutral-100': !!$x.query.searchBox }"
+  <div
+    class="x-layout-container x-layout-max-width-md x-layout-min-margin-16 tablet:x-layout-min-margin-24"
```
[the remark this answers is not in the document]
ajperezbau: The recommendations also have too much margin.

src/components/predictive-layer/predictive-layer.vue
```diff
@@ -1,198 +1,184 @@
 <template>
```
ajperezbau, line 1: There is a glitch when opening the `My History` aside on mobile.

https://user-images.githubusercontent.com/75546736/219642495-ddf3ba44-b7d1-43db-bb9d-242e62f6a6b6.mov

ajperezbau, line 1: I think the remove buttons should be aligned with the clear input button.

image

src/App.vue
```diff
@@ -41,7 +41,9 @@
   export default class Layer extends Vue {
     protected breakpoints: Dictionary<number> = {
       mobile: 743,
-      desktop: Number.POSITIVE_INFINITY
+      tablet: 1439,
```
ajperezbau, line 44: Related tags in the tablet view.

image

ajperezbau: Things like this happen in multiple sites, like the selected filters.

ajperezbau (changes requested):
Good job 👌

src/App.vue
```diff
@@ -41,7 +41,9 @@
...
+      desktop: 1919,
+      large: Number.POSITIVE_INFINITY
```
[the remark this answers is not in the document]
ajperezbau: An example of what I am talking about:

image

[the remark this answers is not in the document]
tajespasarela: fixed

src/App.vue
tajespasarela: that is ok, as it is the mobile layout

src/components/mobile/mobile.vue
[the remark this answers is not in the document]
tajespasarela: fixed

src/components/predictive-layer/predictive-layer.vue
CachedaCodes: This is the empathize appearing above the modal until it collapses. Giving the modal a `z-index` of around 10 fixes it (don't know if that would be the best solution)

CachedaCodes: This is because of the `x-suggestion-group` XDS component and how it styles the button (aspect ratio of 1 and inherits the min height). We fix this here or make changes to the component? (voting for changing the component, maybe a new variable, like we did for the `x-button` for similar problems)

CachedaCodes (changes requested):

src/components/predictive-layer/predictive-layer.vue
tajespasarela: fixed

src/components/desktop/desktop.vue
```diff
@@ -1,94 +1,90 @@
...
           </div>
           <LocationProvider location="predictive_layer">
             <RelatedTags v-if="$x.relatedTags.length > 0" />
           </LocationProvider>
         </div>
 
-        <div
-          class="x-row__item x-row__item--span-2 x-list x-list--horizontal x-list--justify-end x-list--wrap-reverse"
-        >
+        <div class="x-flex x-flex-row x-justify-end">
```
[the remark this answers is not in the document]
tajespasarela: fixed

src/components/desktop/desktop.vue
```diff
@@ -101,33 +97,35 @@
...
       BaseIdModal,
-      CartIcon,
       CloseMainModal,
       CrossIcon,
-      CrossTinyIcon,
-      DesktopToolbar,
       DesktopMyHistoryAside,
-      Empathize,
+      DesktopToolbar,
       FixedHeaderAndAsidesLayout,
```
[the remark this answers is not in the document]
tajespasarela: done

src/components/predictive-layer/predictive-layer.vue
tajespasarela: fixed with the solution @CachedaCodes proposed

herrardo:
@cris-crash we need a token for the aside please :)

CachedaCodes (dismissed):

mnavarroespinosa (approved):

src/components/predictive-layer/predictive-layer.vue
```diff
@@ -1,201 +1,184 @@
...
-        :animation="suggestionsAnimation"
-        class="x-row__item x-row__item--span-12 x-list x-list-horizontal x-padding--05"
-        :class="$x.device === 'mobile' ? 'x-list--gap-03 x-padding--top-00' : 'x-list--gap-02'"
-      >
-        <BaseResultLink
-          v-slot="{ result }"
-          :result="identifierResult"
-          class="x-suggestion-lg x-suggestion desktop:x-suggestion-md"
+  <Empathize :animation="empathizeAnimation" class="x-bg-neutral-0 desktop:x-py-16 desktop:x-pl-16">
+    <component :is="$x.device === 'desktop' ? 'div' : 'BaseScroll'" class="x-h-full">
```
tiborux, line 3: With this change, the empathize will paint a white block:

image

tajespasarela: true, we have to create 2 different components for predictive layer, one for Mobile and another for Desktop

mnavarroespinosa: The white block is displayed because of the padding that is added no matter if the Empathize has content or not, so we could add it conditionally. 

```
    :class="{ 'desktop:x-py-16 desktop:x-pl-16': showEmpathize }"
```

Anyways, it is only happening when there are neither Query Suggestions nor Next Queries, otherwise, the Next Queries will be displayed:

image

I was thinking that, in the scenario of no query suggestions, maybe we should display Top Clicked and Popular Searches instead of not showing anything at all in the predictive. Because not showing at all is causing this empty screen (until enter key is pressed):

image

tajespasarela: Yes, the space is because of padding. But avoiding padding is easily "breakable" on styles change during setup.

The empathize controls if something is rendered inside or not, and if does have not anything inside then it is not rendered.

The `<component>` there making this problem is because we want different components and behaviour on mobile and desktop.

That's the source of the problem we have to solve.

Changing the content to avoid empty case, is not also a solution as also is easily "breakable" on a setup.

mnavarroespinosa: You're right @tajespasarela, avoiding padding is easily "breakable". But I don't get why we need to create two different components for the predictive layer. We just need to render the `<component>` conditionally so the Empathize is not rendered if has no content, right?

tajespasarela: Is something we have to do. It would be better to do on another task. so if you want to solve this in another way, go ahead, but this is something we have to do eventually.

tiborux (approved):

mnavarroespinosa (approved):

PR #266, opened 2023-03-15
feat(design-system): use `tag` from `XDS`

mnavarroespinosa:
EX-8003


ajperezbau (commented):
The predictive components test is failing me in local, but it is working in the CI 🤔 
Good job! 👌

src/components/search/results/next-queries-group.vue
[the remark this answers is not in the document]
mnavarroespinosa: Yes, maybe we should remove this component completely.

tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -8,6 +8,14 @@ Given('an intercepted search response', () => {
   cy.intercept(searchEndpoint).as('interceptedResults');
 });
 
+Given('an intercepted search response from {string}', (origin: string) => {
```
ajperezbau, line 11: I think here we can filter from the origin directly:

```suggestion
Given('an intercepted search response from {string}', (origin: string) => {
  cy.intercept({ url: searchEndpoint, query: { origin: `*${origin}*` } }).as(
    `interceptedResultsFrom:${origin}`
  );
});
```

tests/e2e/cucumber/common-steps.spec.ts
```diff
@@ -52,9 +52,12 @@ When('navigating back', () => {
 });
 
 // Requests
-Then('search request contains the origin {string} in the URL', (origin: string) => {
-  cy.wait('@interceptedResults').its('request.url').should('contain', `origin=${origin}`);
-});
+Then(
```
ajperezbau, line 55: What do you think about changing this to something like:

```suggestion
Then('a search request from {string} is done', (origin: string) => {
  cy.wait(`@interceptedResultsFrom:${origin}`);
});
```

That way we avoid repeating the origin, and the interceptor is filtering it by origin anyway.

ajperezbau (approved):
Nice 👌

tiborux (approved):

PR #267, opened 2023-03-15
feat(design-system): use xds badge

ajperezbau:
The archetype designs are not updated yet, but Alicia said me to use the default badge with the auxiliary color, except in the mobile open aside button, where the lead color is used.

Keep in mind that we will use the badge in the facets headers and the filter counts, which were styled previously with independent classes.

EX-8004


mnavarroespinosa (approved):
lgtm!

tiborux:
In mobile, the button to open the aside is not well styled (size):

image

Staging:

image

CachedaCodes (approved):

tiborux (approved):

PR #273, opened 2023-04-12
feat(facets): select only child filters

ajperezbau:
EMP-265


mnavarroespinosa (changes requested):
For some reason, it is not loading properly the selected filters from the URL when reloading the page.

For example, selecting `Watches` and `Free Gifts`

image

And then, reloading:

image

mnavarroespinosa (approved):

diegopf (approved):

PR #274, opened 2023-04-12
fix: prevent query to be used as extra param

diegopf:
EMP-365


public/snippet-script.js
```diff
@@ -54,6 +54,7 @@ const currency = popFromURLParameters('currency') || 'EUR';
 const consent = popFromURLParameters('consent') !== 'false';
 const documentDirection = popFromURLParameters('doc-dir') || 'ltr';
 const store = popFromURLParameters('store') || undefined;
+const query = popFromURLParameters('query'); // prevent the query from be included as extra param
```
ajperezbau, line 57: We are not using the query anywhere, so I would just pop the parameter without assigning it to a constant.

```suggestion
popFromURLParameters('query'); // prevent the query from be included as extra param
```

ajperezbau (commented):
Nice! 👌

mnavarroespinosa (approved):

tiborux (approved):

PR #276, opened 2023-04-17
feat(design-system): add xds filters

diegopf:
EMP-120


src/components/search/facets/custom-sliced-filters.vue
```diff
@@ -1,13 +1,17 @@
 <template>
   <SlicedFilters v-bind="$attrs">
     <template #show-more>
-      <div class="x-list x-list--horizontal x-list--align-center x-list--gap-03">
+      <div
```
ajperezbau, line 4: I think we can remove this `div` and just place here the icon and the text.

src/components/search/facets/custom-sliced-filters.vue
```diff
@@ -1,13 +1,17 @@
 <template>
   <SlicedFilters v-bind="$attrs">
```
ajperezbau, line 2: I think we should add a dynamic prop to add classes to the `sliced-filters` buttons.
After that, we would just apply the `x-facet-filter-lg` class here.

src/components/search/facets/custom-facets.vue
ajperezbau, line 1: Should we use the rounded badge or the default one in the selected filter count?

image

diegopf: I don't think we have to change anything there, at least for now.

mnavarroespinosa (dismissed):
good job!

ajperezbau (approved):
Good job 👌

mnavarroespinosa (approved):

PR #286, opened 2023-05-18
feat: create a full width empathize component

CachedaCodes:
[EMP-741](https://searchbroker.atlassian.net/browse/EMP-741)

## Motivation and context
We needed to be able to replicate designs that required a full width empathize.
Employing the new layout with customizable alignment to do so. PR started complicating due to the max width layout preventing the overlay of the full width empathize (using a shadow wasn't really a good solution either) of covering the whole width of the screen. Also issues with the z-index of the products in the query previews and the empathize overlay, and then dealing with a lot of duplicated code from the two types of predictive layer. Summarizing:

image

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Replace `DesktopHeaderFloatingPredictive` with `DesktopHeaderFullPredictive` in `desktop`.

[EMP-741]: https://searchbroker.atlassian.net/browse/EMP-741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mavmaf (changes requested):
Please check scroll in mobile

src/components/predictive-layer/predictive-helpers.ts
ajperezbau, line 1: Should we move this to a `composables` folder?

src/components/predictive-layer/full-width-predictive.vue
ajperezbau, line 1: Should we respect the order of the elements stated in the designs?

History Queries | Popular Searches | Next Queries |

https://www.figma.com/file/mYfJZHn668mGEGf61Sji6m/X-Archetype?type=design&node-id=6627%3A128445&t=WyUhqBP93UCfSBNz-1

src/components/predictive-layer/full-width-predictive.vue
```diff
@@ -0,0 +1,225 @@
+<template>
```
ajperezbau, line 1: Should the remove buttons of the `history-queries` be aligned to the right?

image

src/components/predictive-layer/full-width-predictive.vue
ajperezbau, line 1: In the designs, the `lead` color seems to be used for the titles.

src/components/predictive-layer/predictive-identifier-results.vue
```diff
@@ -0,0 +1,44 @@
...
+
+<script lang="ts">
+  import { AnimationProp, BarCodeIcon, BaseResultLink } from '@empathyco/x-components';
+  import { IdentifierResult, IdentifierResults } from '@empathyco/x-components/identifier-results';
+  import { defineComponent } from 'vue';
+
+  export default defineComponent({
+    components: {
+      BarCodeIcon,
+
```
ajperezbau, line 30: Remove these blank lines

src/components/desktop/desktop-header-full-predictive.vue
```diff
@@ -0,0 +1,56 @@
...
+      </header>
+    </div>
+
+    <LocationProvider location="predictive_layer">
+      <FullWidthPredictive />
+    </LocationProvider>
+
+    <div class="desktop-width-item">
+      <div
+        class="x-layout-container x-layout-container-ml-[calc(142px+48px)] x-layout-container-mr-[calc(40px+48px)]"
```
ajperezbau, line 25: I am not sure about placing the related tags with an arbitrary value here, as it can break the alignment if we change the logo or the width of the search input.
Maybe we should create a 3 cols grid to place this items and keep them always aligned.

ajperezbau (changes requested):
Good job! 👌

ajperezbau (changes requested):
Good job! 👌

src/components/desktop/desktop-header-full-predictive.vue
CachedaCodes: How do you align the logo + searchbar + close icon row and the related tags row when they cant be under the same div? The predictive has to be between them because it has to be over the RL when it opens and it also has to be outside a ny div preventing it from being 100% width.

The RL width also has to be limited for the sliding panel to work, so a regular 3 column setup without fixed values for the first and last columns wouldn't work either.

I couldn't find a better solution for this.

src/components/predictive-layer/full-width-predictive.vue
```diff
@@ -0,0 +1,225 @@
...
+      };
+
+      return {
+        isDesktopOrGreater,
+        isTabletOrLess,
+        empathizeAnimation,
+        suggestionsAnimation,
+        showOverlay,
+        handleEmpathizeEvent,
+        ...predictiveHelpers()
```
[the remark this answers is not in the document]
CachedaCodes: I'll do it, but I hate it and I hate Vue Mastery now xD

src/components/predictive-layer/predictive-layer.vue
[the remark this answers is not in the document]
CachedaCodes: I can't reproduce, can you share a video?

CachedaCodes:
> Review the errors logged in the console:
> 
> image

@ajperezbau happens also in main, seems to be related to the currency formatter of X, I'll check a little bit create a task for it.

CachedaCodes:
> GJ! Please check scroll in mobile 💀
> 
> Design related:
> 
> * `remove-history-query` buttons should be aligned to the right
> * Top clicked products. Price is missing.
> * Space between NQs and QSs should be wider?
> * "Clear all" HQ button. Personally, I would change the way is placed (aligning it below `remove-history-query` buttons or copying the way it's displayed on mobile). But if there is a different agreement, then 🤐

The designs for the full with are not final and for whatever reason they couldn't be finished, so here is my thought process:

* Soto is using them like that for the last designs he did so leaving them as is for now
* I didn't touch the top clicked, they are the same as the normal empathize, so for consistency I'm leaving the same for both. If we're adding price to the sliding recommendations then better in another PR and do it for both.
* The structure in that row is a column grid with 4 columns, that spacing is the maximum allowed by the DS for 4 columns I think, when the designs are final and can be refined we'll change it if needed.
* Done

ajperezbau (dismissed):

herrardo (approved):

PR #308, opened 2023-08-02
feat: implement css injector in archetype

CachedaCodes:
Implementation of the css injector in the archetype.

This PR should encapsulate the tailwindcss styles inside the x overlay.

Test it by setting a button with the class `x-button` in the `index.html`. When the injection is applied, it should look like a regular button. When it's not applied, it should have the styles of the X design system.

The injection only works with the build process, so run `build:serve` to see it.


package.json
```diff
@@ -31,7 +31,7 @@
   "dependencies": {
     "@empathyco/x-adapter": "^8.0.0-alpha.22",
     "@empathyco/x-adapter-platform": "^1.0.0-alpha.79",
-    "@empathyco/x-archetype-utils": "^0.1.0-alpha.15",
+    "@empathyco/x-archetype-utils": "file:../x/packages/x-archetype-utils/empathyco-x-archetype-utils-1.0.0-alpha.2.tgz",
```
CachedaCodes, line 34: Change with your generated build file

tests/e2e/support/index.ts
```diff
@@ -108,7 +108,7 @@ const customCommands: CustomCommands = {
       cy.typeQuery(query).type('{enter}');
     });
   },
-  typeQuery: query => cy.getByDataTest('search-input').type(query),
+  typeQuery: query => cy.getByDataTest('search-input').type(query, { force: true }),
```
mavmaf, line 111: needed but not desirable. 
Seems to be working without forcing it in newer Cypress versions https://github.com/cypress-io/cypress/issues/21433

PR #320, opened 2023-09-26
feat: add brand recommendations with filter and extra params

diegopf:
EMP-2200


annacv (approved):
Hi! Reminder: Finally, we agreed in the daily that we would deal with the sort extra-param use case in another task.

herrardo (approved):

PR #326, opened 2023-10-12
feat: apply experience controls for the semantic queries

CachedaCodes:
EMP-2375

## Motivation and context
The new feature of experience controls is already in x-components so we're applying the configurations for the semantic queries.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
You have to be using the port 8080 in your localhost and have the VPN active. Then check a query with semantic queries (battery for example) and check that the configurations from the experience controls endpoint are applied.


annacv (approved):
gj!

PR #397, opened 2023-12-09
i18n message was translated into German

KirilCycle:
json translate was added to message file


herrardo (changes requested):
@KirilCycle Thanks for your contribution!! 

Please add the file to the index.ts file and you are done.

https://github.com/empathyco/x-archetype/blob/main/src/i18n/messages/index.ts

herrardo (approved):
great job!

PR #400, opened 2023-12-14
fix: lazy loading chunks

diegopf:
Recently we discovered an issue with the dynamic chunks being loaded all at once instead of lazy. This is related to an issue on the latest version for [rollup-plugin-generate-html-template ](https://github.com/bengsfort/rollup-plugin-generate-html-template/issues/34).

This PR makes the version stick to previous one until the issue is solved.


annacv (approved):
🎉

PR #439, opened 2024-01-31
feat: implementing use of popular searches as a QP fallback

albertjcuac:
# Pull request template
 
Some customers don’t set brand recommendations. An idea for a fallback is to create a component that takes the popular searches and creates query previews with them, similar to what we do with the semantic queries. 
I have implemented an intermediate component that wraps around custom-query-preview. This component receives an optional prop "maxPopularSearchesToRender", and in the case where QueriesPreview does not exist, it passes as information to be rendered by custom-query-preview the popular searches limited by the aforementioned prop.

Use this in the console to test when there are no queriePreviews: InterfaceX.setSnippetConfig({
queriesPreview: [
]
})
## Motivation and context

- [X] Open issue. If applicable, link: [EMP-3422](https://searchbroker.atlassian.net/browse/EMP-3422?atlOrigin=eyJpIjoiYTkwNDg2YWM2NTlkNDgxNDg0MjkwOThlYWY1MmIzZmEiLCJwIjoiaiJ9)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:

- [ ] My code follows the **[style guidelines](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md#style-guides)** of this project.
- [ ] My changes generate **no new warnings**.
- [ ] I have added **tests** that prove my fix is effective or that my feature works.
- [ ] New and existing **unit tests pass locally** with my changes.

[EMP-3422]: https://searchbroker.atlassian.net/browse/EMP-3422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,58 @@
...
+  import { QueryPreviewInfo } from '@empathyco/x-components/queries-preview';
+  import { Suggestion } from '@empathyco/x-types';
+  import CustomQueryPreview from './custom-query-preview.vue';
+
+  export default defineComponent({
+    components: {
+      CustomQueryPreview
+    },
+    props: {
+      maxPopularSearchesToRender: {
```
lauramargar, line 25: Add a `default` value in this prop, and I think you can remove `required` field. For example:
```
maxPopularSearchesToRender: {
    type: number,
    default: 4
}
```

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,58 @@
...
+          | QueryPreviewInfo[]
+          | {
+              value: QueryPreviewInfo[];
+            }
+        >('queriesPreviewInfo', []);
+
+        return 'value' in injectedQueriesPreview
+          ? injectedQueriesPreview.value
+          : injectedQueriesPreview;
+      });
```
lauramargar, line 42: You can simplify with this:
```suggestion
      const injectedQueriesPreview = inject<{ value: QueryPreviewInfo[] }>('queriesPreviewInfo', { value: [] });
      const queriesPreviewInfo = computed<QueryPreviewInfo[]>(() => injectedQueriesPreview.value);
```

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,58 @@
...
+
+      const hasQueryPreviews = computed<boolean>(() => queriesPreviewInfo.value.length !== 0);
+
+      const getQueriesPreviewInfo = (info: readonly Suggestion[]): QueryPreviewInfo[] => {
```
lauramargar, line 46: Rename this variable
```suggestion
      const getQueriesPreviewInfo = (suggestion: readonly Suggestion[]): QueryPreviewInfo[] => {
```

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,58 @@
...
+        const queryPreviewInfo: QueryPreviewInfo[] = info.map(item => ({
+          query: item.query
+        }));
+        return props.maxPopularSearchesToRender
+          ? queryPreviewInfo.slice(0, props.maxPopularSearchesToRender)
+          : queryPreviewInfo;
```
lauramargar, line 52: If this method is only for `popularSearches`, I think you don't need to make this condition. In my opinion, they should always be limited.

src/components/desktop/desktop.vue
```diff
@@ -14,7 +14,7 @@
           </div>
 
           <LocationProvider location="no_query">
-            <CustomQueryPreview class="x-mt-56" />
+            <ManagePreSearch :maxPopularSearchesToRender="5" class="x-mt-56" />
```
lauramargar, line 17: We also need this in mobile

herrardo:
Like it!

lauramargar (dismissed):

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,48 @@
...
+      maxPopularSearchesToRender: {
+        type: Number,
+        default: 4
+      }
+    },
+    setup(props) {
+      const injectedQueriesPreview = inject<{ value: QueryPreviewInfo[] }>('queriesPreviewInfo', {
+        value: []
+      });
+      const queriesPreviewInfo = computed<QueryPreviewInfo[]>(() => injectedQueriesPreview.value);
```
CachedaCodes, line 34: This can be just one computed. Also, we have to be able to handle the future change to vue 2.7 provides, so handle the unwrapping of the `value` for that case:

```suggestion
      const injectedQueriesPreviewInfo = computed<QueryPreviewInfo[]>(() => {
        const injectedQueriesPreview = inject<QueryPreviewInfo[] | { value: QueryPreviewInfo[] }>(
          'queriesPreviewInfo',
          []
        );

        return 'value' in injectedQueriesPreview
          ? injectedQueriesPreview.value
          : injectedQueriesPreview;
      });
```

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,48 @@
...
+
+      const hasQueryPreviews = computed<boolean>(() => queriesPreviewInfo.value.length !== 0);
+
+      const getQueriesPreviewInfo = (suggestion: readonly Suggestion[]): QueryPreviewInfo[] => {
+        const queryPreviewInfo: QueryPreviewInfo[] = suggestion.map(item => ({
+          query: item.query
+        }));
+        return queryPreviewInfo.slice(0, props.maxPopularSearchesToRender);
+      };
```
CachedaCodes, line 43: It's difficult to understand what this method is doing. It's called `getQueriesPreviewInfo`, but you already have a computed property called `queriesPreviewInfo`. It's not clear why it's needed or what it does.

I suggest convert it to a computed property, return the injected queriesPreview and if there's none then use the `useStore` composable (define it outside of the computed) to take the popular searches and convert them to queryPreview Info. 

Also, we need to register the popular searches module if we're going to use them.

CachedaCodes (changes requested):

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,64 @@
...
+
+      const infoToRender = computed<QueryPreviewInfo[]>(() =>
+        hasQueryPreviews.value
+          ? injectedQueriesPreviewInfo.value
+          : popularSearchesAsQueryPreviewInfo()
+      );
```
CachedaCodes, line 59: `popularSearchesAsQueryPreviewInfo` should be a computed property (it access the store, so it's dynamic), so you can just combine `popularSearchesAsQueryPreviewInfo` and `infoToRender` into one computed property that returns the info to render, I suggest the name `querysPreviewToRender`.

CachedaCodes (changes requested):

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,64 @@
...
+          ? injectedQueriesPreview.value
+          : injectedQueriesPreview;
+      });
+
+      const hasQueryPreviews = computed<boolean>(
+        () => injectedQueriesPreviewInfo.value.length !== 0
+      );
+
+      const popularSearchesAsQueryPreviewInfo = (): QueryPreviewInfo[] => {
+        XPlugin.registerXModule(popularSearchesXModule);
```
CachedaCodes, line 46: Also, if you convert this to a computed property, do this directly in setup. Ideally this should be called once if the component has to use the popular searches, but it's not critical

CachedaCodes (changes requested):

annacv:
Hi!
I've seen we are changing the "CustomQueryPreview" component's name to "BrandRecommendations" in some customers. Maybe we should also do it here?
cc @CachedaCodes @lauramargar

CachedaCodes:
> Hi! I've seen we are changing the "CustomQueryPreview" component's name to "BrandRecommendations" in some customers. Maybe we should also do it here? cc @CachedaCodes @lauramargar

I wouldn't. The component manage pre search has a good enough name (I would name it `PreSearchManager`) and the custom query previews component is not only being used for brand recommendations (if we understand brand recommendations as the queries that the brand puts in the snippet config).

src/components/pre-search/manage-pre-search.vue
```diff
@@ -0,0 +1,60 @@
...
+          'queriesPreviewInfo',
+          []
+        );
+        return 'value' in injectedQueriesPreview
+          ? injectedQueriesPreview.value
+          : injectedQueriesPreview;
+      });
+
+      const hasQueryPreviews = computed<boolean>(
+        () => injectedQueriesPreviewInfo.value.length !== 0
```
diegopf, line 43: If the user does not provide queryPreviews at all in the snippet this will fail and nothing will be shown. We should be ready for that, since it is precisely the use case we want. In the App.vue you can use an empty array as fallback for the provider.

diegopf (changes requested):

lauramargar (approved):

CachedaCodes (approved):

PR #445, opened 2024-02-07
Change empathize layout for mobile devices

annacv:
[EMP-3460](https://searchbroker.atlassian.net/browse/EMP-3460)

Currently in the archetype for mobile devices, the query previews are shown in the pre-search state. However, once you focus on the search box there is no way to go back and see the query previews since the empathize is always taking the full height of the device.

Changes (only mobile): 
- Removed the `x-layout-stack` class when there is no query, so query previews are still shown below the predictive layer & user can scroll to its content.
- Add spacing below the predictive layer, so the spacing between it & query previews is consistent (visually we have the same space between predictive as the one between each queries preview carousel). 
- Add the `x-mt-8` in popular searches: this was sometimes a recurring change in setups, as the distance between history queries & popular searches block was fewer than between popular searches & top clicked carousel one.

[EMP-3460]: https://searchbroker.atlassian.net/browse/EMP-3460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/mobile/mobile.vue
+++ b/src/components/mobile/mobile.vue
@@ -13,7 +13,9 @@
 
     <div class="x-layout-expand" :class="{ 'x-layout-stack': $x.query.search }">
       <LocationProvider location="predictive_layer" class="x-z-10">
-        <PredictiveLayer :class="{ 'x-mb-56': !$x.query.search }" />
+        <PredictiveLayer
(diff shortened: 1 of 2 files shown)


albertjcuac:
Seems fine to me but maybe it could be good to add something to visually define the end of the empathize, like a border shadow.

albertjcuac (approved):
nice!

PR #492, opened 2024-07-12
fix: Fix backroom back from login context issue

alvarodE:
## Motivation and context
We have a bug in Backroom in which if after performing a query the user logs in, when returning to the Search layer Backroom context is not updated.

This is because the Backroom context was only updated with the `SearchRequestChanged`event but that event is not fired in the first search if loaded from the URL; adding a single line on the `ParamsLoadedFromUrl` fixes the issue.

- [x] Open issue: https://searchbroker.atlassian.net/browse/PBU-31

## Type of change

- [x] `Main`

## Checklist:
- [x] My code follows the **[style guidelines](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md#style-guides)** of this project.
- [x] I have performed a **self-review** on my own code.
- [x] I have **commented** my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate **no new warnings**.
- [x] I have added **tests** that prove my fix is effective or that my feature works.
- [x] New and existing **unit tests pass locally** with my changes.

### DIFF
--- a/public/snippet-script.js
+++ b/public/snippet-script.js
@@ -119,7 +119,7 @@ window.addEventListener('load', () => {
     instance,
     lang,
     audience: 'enterprise',
-    playboardUrl: 'https://playboard-platform.staging.empathy.co/empathy',
+    playboardUrl: `https://playboard-platform.staging.empathy.co/${instance}`,
     appContainerSelector: '.x-root-container',
     searchLayerSelector: '.x'
   });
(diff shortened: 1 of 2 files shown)


public/snippet-script.js
```diff
@@ -116,9 +116,10 @@ window.addEventListener('load', () => {
     analytics: {
       baseUrl: 'https://api.staging.empathy.co/statistics/v2'
     },
-    lang,
     instance,
+    lang,
     audience: 'enterprise',
+    playboardUrl: 'https://playboard-platform.staging.empathy.co/empathy',
```
alvarodE, line 122: Not specifically part of this fix but we also want to include the missing Playboard link in Backroom.

diegopf (approved):

PR #525, opened 2024-12-10
feature: Fix RPs layout for mobile devices

annacv:
# Pull request template

Describe the purpose of the change, the specific changes done in detail, and the issue you have fixed.

## Motivation and context

- [ ] Open issue. If applicable, link:

## Type of change

- [ ] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:

- [ ] My code follows the **[style guidelines](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md#style-guides)** of this project.
- [ ] My changes generate **no new warnings**.
- [ ] I have added **tests** that prove my fix is effective or that my feature works.
- [ ] New and existing **unit tests pass locally** with my changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1229,7 +1229,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
-      "dev": true,
+      "extraneous": true,
       "inBundle": true,
       "license": "MIT",
       "engines": {
--- a/src/components/main.vue
+++ b/src/components/main.vue
@@ -4,10 +4,10 @@
     <LocationProvider location="results">
       <Results />
     </LocationProvider>
-    <LocationProvider v-if="showRelatedPrompts" location="related_prompts">
+    <LocationProvider v-if="showRelatedPrompts" :location="semanticsLocation">
       <CustomRelatedPrompts
         :related-prompt-list="relatedPrompts"
-        :class="x.noResults ? 'desktop:x-mt-24' : 'x-mt-24 desktop:x-mt-32'"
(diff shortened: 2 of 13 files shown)


src/components/main.vue
```diff
@@ -4,10 +4,10 @@
     <LocationProvider location="results">
       <Results />
     </LocationProvider>
-    <LocationProvider v-if="showRelatedPrompts" location="related_prompts">
+    <LocationProvider v-if="showRelatedPrompts" :location="semanticsLocation">
```
victorcg88, line 7: Is this intentional? Is it correct?

package-lock.json
```diff
@@ -1229,7 +1229,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
-      "extraneous": true,
+      "dev": true,
```
victorcg88, line 1232: Why is this being reverted?

victorcg88 (commented):
IMHO, the top and the bottom margins of the prompts could look better if they are the same. WDYT?

image

Also, I'm wondering if we should add still a bit more space to the bottom, so the prompt is clearly visible

image

But this is subjective, maybe is worth to ask the rest of the team.

package-lock.json
annacv: Hi, I thought I had uploaded this file too, sorry

src/components/main.vue
annacv: Yep, related_prompts should be added as a location value in x-components if we want to use it here,
maybe it should also be discussed with the team @victorcg88

victorcg88 (approved):

PR #613, opened 2025-08-12
refactor: migrate to script setup & speedup build

davidglezz:
- Automatic code modernization by copilot
  -  Use script setup → less verbose, better DX 
  -  image
- Build improvements:
  - Do not typecheck on build (with tsc) → Must be done with `vue-tsc`.
  - Do not generate bundle size report. It should be enabled only when used. 99% of the times is not viewed.
  - Build time reduced by 20..25%

### DIFF
--- a/src/components/search/results/results.vue
+++ b/src/components/search/results/results.vue
@@ -47,7 +47,7 @@
   </ResultsList>
 </template>
 
-<script lang="ts">
+<script setup lang="ts">
 import type { RelatedPromptNextQuery } from '@empathyco/x-types'
 import {
   BaseVariableColumnGrid,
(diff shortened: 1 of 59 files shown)


AdrianArenal (approved):

src/components/search/results/results.vue
```diff
@@ -65,67 +65,37 @@ import {
...
+const queriesPreviewInfo = computed(() => {
+  if (relatedPrompts.value.length) {
+    const queries = [] as string[]
+    relatedPrompts.value[selectedPrompt.value].relatedPromptNextQueries?.forEach(
+      (nextQuery: RelatedPromptNextQuery) => queries.push(nextQuery.query),
     )
+    return queries.map(query => ({ query }))
+  }
+  return []
+})
```
davidglezz, line 94: TODO: simplify & extract all occurrences
Should it be a getter?

PR #637, opened 2025-10-07
feat: add next queries cta

victorcg88:
https://searchbroker.atlassian.net/browse/RST-3445


lauramargar:
Great job!! 
Only one thing. If the user closes the next-queries-cta and then clicks again on a CTA button, it adds a white space.
Screenshot 2025-10-08 at 08 50 56

